PR #25, opened 2021-04-27
Implement envfile

### LINKED ISSUE #22: Feature: load .env
`.env` is a file that specifies the environment variables that should be loaded. AFIAK Poetry doesn't load these. Will `poe` do, though?

ThatXliner:
Closes #22

### DIFF
--- a/poethepoet/config.py
+++ b/poethepoet/config.py
@@ -15,6 +15,7 @@ class PoeConfig:
         "default_array_task_type": str,
         "default_array_item_task_type": str,
         "env": dict,
+        "envfile": str,
         "executor": dict,
     }
 
@@ -51,6 +52,10 @@ class PoeConfig:
     def global_env(self) -> Dict[str, str]:
         return self._poe.get("env", {})
 
+    @property
+    def envfile(self) -> str:
+        return self._poe.get("envfile", "")
+    
     @property
     def project(self) -> Any:
         return self._project
--- a/poethepoet/config.py
+++ b/poethepoet/config.py
@@ -15,7 +15,7 @@ class PoeConfig:
         "default_array_task_type": str,
         "default_array_item_task_type": str,
         "env": dict,
-        "envfile": str,
+        "envfile": dict,
         "executor": dict,
     }
 
@@ -53,9 +53,11 @@ class PoeConfig:
         return self._poe.get("env", {})
 
     @property
-    def envfile(self) -> str:
-        return self._poe.get("envfile", "")
-    
+    def envfile(self) -> Dict[str, str]:
+        if self._poe.get("envfile"):
+            return {name: contents for line in Path(self._poe["envfile"]).resolve().read_text().splitlines() for name, contents in line.split("=", 1)}
+        return {}
+
     @property
     def project(self) -> Any:
         return self._project
@@ -117,6 +119,13 @@ class PoeConfig:
             )
         # Validate env value
         env = self.global_env
+        if env:
+            for key, value in env.items():
+                if not isinstance(value, str):
+                    raise PoeException(
+                        f"Value of {key!r} in option `env` should be a string, but found {type(value)!r}"
+                    )
+        env = self.envfile
         if env:
             for key, value in env.items():
                 if not isinstance(value, str):
--- a/poethepoet/config.py
+++ b/poethepoet/config.py
@@ -56,7 +56,7 @@ class PoeConfig:
     def envfile(self) -> Dict[str, str]:
         if self._poe.get("envfile"):
             envfile_path = Path(self._poe["envfile"]).resolve()
-            if envfile_path.exists() and envfile_path.is_file():
+            if envfile_path.is_file():
                 return {name: contents for line in envfile_path.read_text().splitlines() for name, contents in line.split("=", 1)}
         return {}
 
--- a/poethepoet/config.py
+++ b/poethepoet/config.py
@@ -55,7 +55,9 @@ class PoeConfig:
     @property
     def envfile(self) -> Dict[str, str]:
         if self._poe.get("envfile"):
-            return {name: contents for line in Path(self._poe["envfile"]).resolve().read_text().splitlines() for name, contents in line.split("=", 1)}
+            envfile_path = Path(self._poe["envfile"]).resolve()
+            if envfile_path.exists() and envfile_path.is_file():
+                return {name: contents for line in envfile_path.read_text().splitlines() for name, contents in line.split("=", 1)}
         return {}
 
     @property
@@ -127,7 +129,7 @@ class PoeConfig:
                     )
         env = self.envfile
         if env:
-            for key, value in env.items():
+            for key, value in env.items():   # TODO: Validate it some other way. Give a warning instead?
                 if not isinstance(value, str):
                     raise PoeException(
                         f"Value of {key!r} in option `env` should be a string, but found {type(value)!r}"
--- a/poethepoet/task/base.py
+++ b/poethepoet/task/base.py
@@ -136,8 +136,9 @@ class PoeTask(metaclass=MetaPoeTask):
         """
         Run this task
         """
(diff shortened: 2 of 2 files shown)


nat-n (commented):
Thanks for picking this up :)

poethepoet/config.py
```diff
@@ -51,6 +52,10 @@ def default_array_item_task_type(self) -> str:
     def global_env(self) -> Dict[str, str]:
         return self._poe.get("env", {})
 
+    @property
+    def envfile(self) -> str:
+        return self._poe.get("envfile", "")
```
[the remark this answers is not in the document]
ThatXliner: What if we do the parsing and everything in here and treat this like `global_env`?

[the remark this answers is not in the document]
ThatXliner: I'm going to try that, actually.

ThatXliner:
Ok, so I *think* I have an implementation working, which I haven't tested yet.

poethepoet/task/base.py
```diff
@@ -136,7 +136,7 @@ def run(
         """
         Run this task
         """
-        env = dict(env or {}, **self._config.global_env)
+        env = dict(env or {}, **self._config.global_env, **self._config.envfile)
```
nat-n, line 139: I believe this will blow up if the same variable is defined both in the global options and in the dotenv file.

poethepoet/config.py
```diff
@@ -51,6 +52,12 @@ def default_array_item_task_type(self) -> str:
     def global_env(self) -> Dict[str, str]:
         return self._poe.get("env", {})
 
+    @property
+    def envfile(self) -> Dict[str, str]:
+        if self._poe.get("envfile"):
```
nat-n, line 57: I'm not 100% sure whether the config is the best place for this. It might be, but it depends on whether we think of the dotenv file as an extension of the toml config, or more a feature of the environment (i.e. context).
 
At any rate it needs to be cached so we don't needlessly open and parse the file multiple times.

poethepoet/config.py
```diff
@@ -112,6 +119,13 @@ def validate(self):
             )
         # Validate env value
         env = self.global_env
+        if env:
+            for key, value in env.items():
+                if not isinstance(value, str):
+                    raise PoeException(
+                        f"Value of {key!r} in option `env` should be a string, but found {type(value)!r}"
+                    )
+        env = self.envfile
```
nat-n, line 128: I'm not sure this is useful as it is here. Parsed values will always be of type string. Also the error message would be wrong (refers to the env option), and if the file doesn't exist we'll get an unhandled exception.

nat-n (commented):
Do you intend to implement support for the task specific envfile option also?

The new functionality will need to be covered by feature tests before it can be included in a release.

poethepoet/task/base.py
ThatXliner: No, one of them will overwrite. Not sure what you mean by "blow up"

poethepoet/config.py
ThatXliner: True. Do you have a cache system already? Or do we need to implement one from scratch?

poethepoet/config.py
ThatXliner: Oh right. I'll get that fixed

ThatXliner: > if the file doesn't exist we'll get an unhandled exception.

Fixed that in the latest commit.

Not sure if we should raise and error if it doesn't exist.

poethepoet/task/base.py
nat-n:
```python
>>> dict({}, **({"foo": "bar"}), **({"foo": "baz"}))
  File "<stdin>", line 1, in <module>
TypeError: type object got multiple values for keyword argument 'foo'
```
Using the update method would be more robust. That said I think it might more sense for the more local variables (in the global options) to take precedence.

poethepoet/config.py
nat-n: A dict attribute on the host object keyed by file path like `envfile_cache: Dict[str, Any]` would be an adequate cache.

I think the ideal logic for loading the files would be something like:
 1. attempt to explicitly check if the file exists, if not then the result is an empty dict (which is cached).
 2. If the file does exist then attempt to read and parse it
 3. for parsing I think it's better to be a bit stricter about the format and use a regex along the lines of `\s*(?P<var>[a-zA-Z_][a-zA-Z_0-9]*)\s*=\s(?P<value>(?:[^\"\'\s]+|\".+\"|\'.+\'))\s*$`. Also ignoring `#` commented lines. Files that fail at this should result in an ValueError.

I think keeping the location of the dotenv file configurable like it is makes the most sense. Accepting `true` instead of 
".env" for the common case is kind of elegant, but on the other hand it also adds a little of under the hood and API complexity, it's a little obscure, and doesn't really simplify much for the user

In case it wasn't clear; my point about whether this logic should belong to the PoeConfig class or the RunContext class, is a question of the conceptual purity of these two classes. The PoeConfig class owns the config from pyproject.toml. The RunContext class owns features of the environment that don't really belong to the config, or any particular class. Maybe dotenv files are better aligned with content...

What do you think?

poethepoet/config.py
nat-n: Ideally the validate method should return error messages rather than raising errors.

poethepoet/config.py
```diff
@@ -51,6 +52,14 @@ def default_array_item_task_type(self) -> str:
     def global_env(self) -> Dict[str, str]:
         return self._poe.get("env", {})
 
+    @property
+    def envfile(self) -> Dict[str, str]:
+        if self._poe.get("envfile"):
+            envfile_path = Path(self._poe["envfile"]).resolve()
+            if envfile_path.exists() and envfile_path.is_file():
```
nat-n, line 59:
```suggestion
            if envfile_path.is_file():
```
Unless I'm missing something?

ThatXliner: oh, I've always wondered about that. if `env_path.exist()` is unnecessary, then sure!

poethepoet/task/base.py
ThatXliner: >  That said I think it might more sense for the more local variables (in the global options) to take precedence.

You mean,

envfile overwrites local overwrites global

```
envfile > local > global
```

nat-n: I mean `self._config.global_env` overwrites `self._config.envfile`. So the most explicit/local source takes precedence.

ThatXliner: There will be global envfile and normal envfile? Just treat it like env but there will be parsing, etc.

In other words, the implementation should be similar to env except

1. it’s not a dict; it's a string signifying the path to the envfile
2. Parsing will be done.

nat-n: In case it isn't clear, `self._config.global_env` is populated in the toml file like so:

```toml
[tool.poe.env]
VAR1 = "foo"
VAR2 = "foo"
```

These values should override any external dotenv file, not the other way around.

hence:
```python
env = dict(env or {})
env.update(self._config.envfile)
env.update(self._config.global_env)
```

poethepoet/config.py
ThatXliner: I have an idea: we default the `envfile` config to `.env` and load it every time. The logic in parsing the envfile should belong to the `PoeConfig` because that way, we can just copy the implementation of the `env` config and act like it.

nat-n:
@ThatXliner Thanks again for taking the time to work on this. I'm afraid it wasn't taking shape the way I hoped, so you'll notice I opened another PR with my own efforts to address the same issue at #29 I hope this isn't a disappointment for you.

I decided it was best to write the envfile parsing logic from scratch, keeping it close to bash syntax. I investigated python-dotenv as an option, but was put off by the performance (not terrible but measurable overhead, since it does more than presently necessary), as well as hesitancy to add another dependency.

ThatXliner:
That's fine. I'm going to close this for now

PR #29, opened 2021-05-25
Support envfile task level and global option

nat-n:
Adds support for a new `envfile` option at global or task level. #22 

Includes an original lightweight parser for envfiles closely following bash syntax.

### DIFF
--- a/poethepoet/helpers.py
+++ b/poethepoet/helpers.py
@@ -2,4 +2,4 @@ import re
 
 
 def is_valid_env_var(var_name: str) -> bool:
-    return bool(re.match("[a-zA-Z_][a-zA-Z0-9_]*", var_name))
+    return bool(re.match("^[a-zA-Z_][a-zA-Z0-9_]*$", var_name))
--- a/tests/unit/test_parse_env_file.py
+++ b/tests/unit/test_parse_env_file.py
@@ -1,0 +1,133 @@
+from poethepoet.envfile import parse_env_file
+import pytest
+
+valid_examples = [
+    (
+        """
+    # empty
+    """,
+        {},
+    ),
+    (
+        """
+        # single word values
+        WORD=something
+        WORD_WITH_HASH=some#thing
+        NUMBER=0
+        EMOJI=😃😃
+        DOUBLE_QUOTED_WORD="something"
+        SINGLE_QUOTED_WORD='something'
+        """,
+        {
+            "WORD": "something",
+            "WORD_WITH_HASH": "some#thing",
+            "NUMBER": "0",
+            "EMOJI": "😃😃",
+            "DOUBLE_QUOTED_WORD": "something",
+            "SINGLE_QUOTED_WORD": "something",
+        },
+    ),
+    (
+        """
+        # multiword values
+        WORD=some\\ thing                   # and trailing comments
+        DOUBLE_QUOTED_WORD="some thing"
+        SINGLE_QUOTED_WORD='some thing'
+        """,
+        {
+            "WORD": r"some thing",
+            "DOUBLE_QUOTED_WORD": "some thing",
+            "SINGLE_QUOTED_WORD": "some thing",
+        },
+    ),
+    (
+        """
+        # values with line breaks
+        WORD=some\\
+thing
+        DOUBLE_QUOTED_WORD="some
+        thing"
+        SINGLE_QUOTED_WORD='some
+        thing'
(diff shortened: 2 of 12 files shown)


poethepoet/helpers.py
```diff
@@ -1,5 +1,128 @@
...
+                state = ParserState.IN_SINGLE_QUOTE
+
+            elif match_terminator == '"':
+                cursor += 1
+                state = ParserState.IN_DOUBLE_QUOTE
+                continue
+
+            else:
+                # We found one or more backslashes
+                num_backslashes = len(match_terminator)
```
ThatXliner, line 82: I see what you're doing. But won't the regex
```regex
((?:".*?(?<!\\)(\\\\)*?")|(?:'.*?(?<!\\)(\\\\)*?'))
```
Be sufficient?

### Explanation

 - 2 groups: one for handling `"` and one for handling `'`.
 - The middle part (`.*?`) should be self-explanatory
 - The last part (`(?<!\\)(\\\\)*?`) will match an even number of `\` that are not escaped.

tests/unit/test_parse_env_file.py
```diff
@@ -0,0 +1,71 @@
...
+        },
+    ),
+    (
+        """
+    # without linebreak between vars
+    FOO=BAR BAR=FOO
+    """,
+        {"FOO": "BAR", "BAR": "FOO"},
+    ),
+]
```
ThatXliner, line 66: Can we add a test for space before the `FOO`, too? And, instead of strictly complying, why not allow spaces between the `=`?

ThatXliner (commented):
Nice. This is tested to work, right?

EDIT: Also, why did you "UI" so much?

tests/unit/test_parse_env_file.py
nat-n: Look again, preceding space is present in all the tests 💯  Maybe there should be a test *without* preceding space...

I considered allowing spaces around the `=` though felt that sticking to how bash does it was a safer bet design wise. I figured it would be least surprising if the file is loaded exactly the same way as it would be via the source command. Maybe this is worth revisiting, though I would need evidence to motivate a change.

nat-n:
Thanks for taking a look.

> Nice. This is tested to work, right?
> 
> EDIT: Also, why did you "UI" so much?

I tried it with a few simple scenarios, I intend it test it more and also add a few scenarios to the feature test suite before merging and making a pre-release.

The "UI" line changes are all about being able to print a warning at context.py:90 when the UI is in verbose mode.

poethepoet/helpers.py
nat-n: Hmm, it seemed simpler this way, but that's worth exploring. Note that backslashes don't works as escapes within single quotes, and we need remove active quotes from the result but preserve escaped backslashes as in lines 84 & 120.

ThatXliner:
The current envfile parsing is very fault tolerant, which is good, but doesn't emit warnings for syntax errors.

With this envfile:
```bash
TEST=yeetus  # Epic


 e = A
 A = b
a = b
a = c
```
Only `TEST` is set. It doesn't send an error! (Removing spaces around the `=` will make it set all of these). Poe doesn't warn on duplicate keys either.

nat-n:
> The current envfile parsing is very fault tolerant, which is good, but doesn't emit warnings for syntax errors.

Good catch, regexes are tricky beasts. I added some tests with bad input to give more confidence.

ThatXliner:
Good, it's erroring now. Except the error message is very cryptic... You might wanna use something like [Lark](https://github.com/lark-parser/lark) which has very epic error messages. (Lark [can generate a standalone](https://lark-parser.readthedocs.io/en/latest/tools.html#stand-alone-parser) so you don't need to worry about an extra dependency)

nat-n:
> Good, it's erroring now. Except the error message is very cryptic... You might wanna use something like [Lark](https://github.com/lark-parser/lark) which has very epic error message support. (Lark [can generate a standalone](https://lark-parser.readthedocs.io/en/latest/tools.html#stand-alone-parser) so you don't need to worry about an extra dependency)

Good point. I tweaked the error handling so it's a bit more specific and helpful I think.

I think it's good to go now, but I'll leave this open a short while longer in case you'd like to give more feedback before I merge and create another pre-release.

ThatXliner:
Looks good. It would be nice to have suggestions but at least the errors are correct.

PR #32, opened 2021-07-03
Initial implementation of Poetry plugin

### LINKED ISSUE #14: Poetry plugin
The roadmap for poetry includes a plugin system https://github.com/python-poetry/poetry/pull/1237 which should allow for poe to run inside the poetry CLI.

Creating a Poe the Poet plugin for poetry would ideally entail that the exact same config in the pyproject.toml could be used to the same effect either via the existing poe CLI, or via poe running as a poetry plugin and invoked via the the poetry CLI.

It'll probably look something like:

```bash
poetry poe test
```

which would be functionally equivalent to the following but with a shorter command a perhaps a minor performance boost:

```bash
poetry run poe test
```

ThatXliner:
Closes #14


ThatXliner:
Should we notify if a command tries to/could've override Poetry default commands?

nat-n:
Interesting. Thanks for making this PoC. I thought it would be more complex (probably the plugin API I looked at is no longer current).

I'll need to do quite a bit of testing before merging/releasing this (and I'm afraid I have other stuff to test first; currently I don't have a lot of time for this :( ), but I'm glad to see it's not too complicated to get working.

I'm of two minds about whether it's better to namespace poe tasks (i.e. `poetry poe test` vs `poetry test`). Though maybe offering a plugin for each and letting the user decide would be reasonable. At least notifying the user if they've configured colliding task name would be in keeping with current design philosophy.

I'd prefer not to keep poetry as an extra dependency (i.e. installed via `pip install poethepoet[poetry_plugin]` or so), because I've found poe is actually useful in contexts without poetry.

The functionality you've demonstrated here _might_ be enough for a first release that adds value, but ultimately some thought needs to go into how to best integrate poe CLI functionality with the poetry CLI, for consistently displaying help/errors etc. This is were I imagined most of the work/refactoring will be required, because poe doesn't currently use cleo.

ThatXliner:
> I'm of two minds about whether it's better to namespace poe tasks (i.e. `poetry poe test` vs `poetry test`). Though maybe offering a plugin for each and letting the user decide would be reasonable. At least notifying the user if they've configured colliding task name would be in keeping with current design philosophy.

I realized that we probably can move this plugin functionality to a different PyPi package. I'm pretty sure when someone does pip install poethepoet`, poethepoet is available as a library (due to Poetry's auto detection I think).

> I'd prefer to keep poetry as an _extra_ dependency (i.e. installed via `pip install poethepoet[poetry_plugin]` or so), because I've found poe is actually useful in contexts without poetry.

Check out my latest commit. I did more than just edit the `pyproject.toml` file.

> The functionality you've demonstrated here _might_ be enough for a first release that adds value, but ultimately some thought needs to go into how to best integrate poe CLI functionality with the poetry CLI, for consistently displaying help/errors etc. This is were I imagined most of the work/refactoring will be required, because poe doesn't currently use cleo.

I see. Can't we just redirect subprocess's stdout and stderr some how?

ThatXliner:
> I see. Can't we just redirect subprocess's stdout and stderr some how?

I did that in the latest commit (not so sure if I should use `copy.copy` or that the variable already copies the value)

ThatXliner:
@nat-n can you review this?

nat-n:
Hi @ThatXliner, Sorry for the delay. I will come back to this, it's an important feature. I'm afraid I don't expect to have time to give it enough focused attention for at least a couple more weeks. I'm afraid another project has been taking all my time for side projects.

ThatXliner:
I guess I could make a plugin separately, as I said before…

nat-n:
Hi @ThatXliner, sorry again for the delay. If you're dying to ship this then indeed it might be best to build it as a separate project.

I absolutely want to have poetry plugin support built into poethepoet. However by current my assessment it's not a quick win; I need to take some time to get a deep understanding of the plugin architecture and conventions in poetry, and think about how to integrate it properly. Meanwhile my highest priority for this project at the moment is to stabilise the major new features that are currently in pre-release on the development branch so that they can be released as 0.11.0, and this might still take a while.

That said it might be useful, when I'm ready to work on this, to have an example to base it on that's already been battle tested.

nat-n:
Also if you maintain this PR then please target the development branch, as per the [contributing guidelines](https://github.com/nat-n/poethepoet/blob/main/.github/CONTRIBUTING.md)

pyproject.toml
```diff
@@ -12,6 +12,10 @@ homepage =  "https://github.com/nat-n/poethepoet"
 python = "^3.6"
 pastel = "^0.2.0"
 tomli = "^1.0.0"
+poetry = {version = "^1.1.7", allow-prereleases = true, optional = true}
```
nat-n, line 15:
```suggestion
poetry = {version = "^1.0", allow-prereleases = true, optional = true}
```
The example from the docs suggests using `"^1.0"` is there any reason to be more prescriptive?

poethepoet/plugin.py
```diff
@@ -0,0 +1,77 @@
...
+        stderr=stderr,
+        check=False,
+    ).returncode
+
+
+class PoePlugin(ApplicationPlugin):
+    def activate(self, application: app.Application) -> None:
+        if app.Application is None:
+            print(
+                "Poetry not found. Did you install this plugin via `poetry plugin add poethepoet[poetry_plugin]`?"
```
nat-n, line 61: When can this happen? Is there a known scenario that something other than poetry would attempt to activate this plugin.

poethepoet/plugin.py
```diff
@@ -0,0 +1,77 @@
...
+    )
+
+from . import config
+
+
+def get_poe_task_names() -> Iterator[str]:
+    return (
+        name
+        for name in tomlkit.loads(
+            Path(config.PoeConfig().find_pyproject_toml()).read_text()
```
nat-n, line 33: I haven't tried it but looking through the code it looks like there's a way to get the pyproject.toml content loaded by poetry without loading it again. Something like `app.config.config`

poethepoet/plugin.py
```diff
@@ -0,0 +1,77 @@
...
+    from cleo.commands.command import Command
+    from poetry.console import application as app
+    from poetry.plugins.application_plugin import ApplicationPlugin
+except ModuleNotFoundError:
+    # Hacky defaults to
+    # tell whether poetry is installed or not,
+    # please mypy,
+    # and please Poetry's plugin system (simulate the ApplicationPlugin object)
+    app = type("application", (), {"Application": None})  # pylint: disable=C0103
+    ApplicationPlugin = type(
```
nat-n, line 20: It looks like this try/except is just about making this module importable other than by poetry. Do you have a use case in mind for this?

nat-n (commented):
I've finally gotten around to looking into this. Sorry for the delay. I anticipate having some time over the next few weeks to get it done (it's the highest priority major feature now after #45).

To start with I'd like your help to understand what you've done so far, to see what's there and what's missing.

poethepoet/plugin.py
ThatXliner: Good point. This module shouldn't be imported unless as a poetry plugin. Unless there is some wanted flexibility in the future? (But I guess we can always add it back)

poethepoet/plugin.py
```diff
@@ -0,0 +1,77 @@
...
+class PoePlugin(ApplicationPlugin):
+    def activate(self, application: app.Application) -> None:
+        if app.Application is None:
+            print(
+                "Poetry not found. Did you install this plugin via `poetry plugin add poethepoet[poetry_plugin]`?"
+            )
+            sys.exit(1)
+        io = application.create_io()
+        # Create seperate commands per task
+        for task_name in get_poe_task_names():
```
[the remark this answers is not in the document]
ThatXliner: "Does it fail?" yes. I have tried it.

poethepoet/plugin.py
```diff
@@ -0,0 +1,77 @@
...
+
+def run_poe_task(name: str, io) -> int:
+    io.fileno = sys.stdin.fileno  # Monkey patches io because subprocess needs it
+    stdin = io
+    io.fileno = sys.stdout.fileno
+    stdout = io
+    io.error_output.fileno = sys.stdout.fileno
+    stderr = io.error_output
+
+    return subprocess.run(
```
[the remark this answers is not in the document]
ThatXliner: We can do that? I've never knew.

My thought was that `poetry X` was just a simple shorthand for `poe X`

poethepoet/plugin.py
```diff
@@ -0,0 +1,77 @@
...
+        for name in tomlkit.loads(
+            Path(config.PoeConfig().find_pyproject_toml()).read_text()
+        )["tool"]["poe"]["tasks"].keys()
+        # Make sure it doesn't override default poetry commands
+        if name not in app.COMMANDS
+    )
+
+
+def run_poe_task(name: str, io) -> int:
+    io.fileno = sys.stdin.fileno  # Monkey patches io because subprocess needs it
```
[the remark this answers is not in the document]
ThatXliner: Well, I said something [here](https://github.com/nat-n/poethepoet/pull/32#issuecomment-873678025)

> >The functionality you've demonstrated here might be enough for a first release that adds value, but ultimately some thought needs to go into how to best integrate poe CLI functionality with the poetry CLI, for consistently displaying help/errors etc. This is were I imagined most of the work/refactoring will be required, because poe doesn't currently use cleo.
>
> I see. Can't we just redirect subprocess's stdout and stderr some how?

poethepoet/plugin.py
ThatXliner: You have a good point. See also https://github.com/nat-n/poethepoet/pull/32#discussion_r774239496

poethepoet/plugin.py
ThatXliner: Interesting. But I wouldn't use anything undocumented. Made a discussion [here](https://github.com/python-poetry/poetry/discussions/4930)

poethepoet/plugin.py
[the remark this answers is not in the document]
nat-n: I see. Maybe the optimal solution would be attempt to use that undocumented API, and fallback to loading the config again upon failure.

nat-n:
I had a quick play around with this implementation and made a few changes in a new PR. #46

If you'd like to continue to experiment and share ideas on this topic then we can keep this PR open, but id's suggest rebasing it onto the other way and updating it to target the branch of the other one. Otherwise we should close this one now. Or it might be easier to create a new PR of that nature since the rebase (which includes everything from the development branch) is non-trivial.

Thanks again for your work on this.

ThatXliner:
Rebased... I think

ThatXliner:
Argh I dislike git

ThatXliner:
I'm going to close this for now:

1. I hate git.
2. I like your progress on the other branch

PR #34, opened 2021-08-21
Allow changing default verbosity.

Artemis21:
For example, setting:

```toml
[tool.poe]
verbosity = -1
```

Will make all commands quiet unless --verbose is passed.


Artemis21:
Three tests fail and I get two type checking errors, but that also happens on the previous commit so I'm assuming it's not something I've done.

poethepoet/__version__.py
```diff
@@ -1 +1 @@
-__version__ = "0.11.0b3"
+__version__ = "0.12.0b0"
```
nat-n, line 1:
```suggestion
__version__ = "0.11.0b4"
```
We're still building up to 0.11.0 :(

nat-n (changes requested):
Hi @Artemis21, thanks for the contribution. Nicely done.

I think it would make sense to also support this option at the task level, though it's not necessary for now.

Fortunately all tests pass in CI :) Out of curiosity, which ones fail for you?

I'm happy to merge and create another prerelease for this, if you could just update the version number as advised. 0.11.0 is set to be a big one, I just need to find time to finish testing of some of the prerelease features.

nat-n:
Actually never mind, I fixed the version myself. Et voila https://pypi.org/project/poethepoet/0.11.0b4/

Artemis21:
Cool!
The test output is [here](https://paste.artemisdev.xyz/raw/anayu). The tests that fail are `test_virtualenv_executor_activates_venv`, `test_virtualenv_executor_provides_access_to_venv_content` and `test_detect_venv`, I think it may be because of some messing around I've done with my Python installations recently.

Mypy gives:
```
poethepoet/completion/zsh.py:38: error: syntax error in type comment
poethepoet/completion/zsh.py:47: error: syntax error in type comment
```
Both look fine to me though so not sure why.

Artemis21:
> I think it would make sense to also support this option at the task level, though it's not necessary for now.

I'll have a look at implementing that. Another nice feature would be the ability to set normal verbosity from the command line - passing neither `-v` or `-q` is no longer equivalent.

The simplest way to do that would be to add a third option, like `--normal-verbosity`. 

A more generic way would be to have an option that allows you to set the level directly, eg. `--verbosity 0`. This would make it easier if more verbosity levels than just `-1`, `0` and `1` were added in the future, but may be a little unintuitive.

An approach users may be more familiar with would be to make `-v` and `-q` simply *adjust* the current verbosity setting by `1` or `-1` respectively. With the default verbosity set to `0`, this would act exactly as it does now, but for example with default verbosity set to `-1`, you could make it verbose with `-vv` or normal with `-v`.

nat-n:
I like the model of `-v` meaning +1 verbosity and `-vv` +2 etc. I considered it initially though wasn't sure how to implement/document it elegantly with argparser, and tbh increasing the verbosity doesn't do much. I had in mind to make it display more detail in help and give more detail about running tasks, but never clarified this into something I was sure would be useful. 

Though adding a `--verbosity` option that accepts a value is probably the most straight forward and robust solution.

PR #35, opened 2021-08-24
Make the -v and -q flags incremental

Artemis21:
See discussion in #34. This means that it is possible to specify normal verbosity again, even if a default verbosity is configured, and will make it easier to support more verbosity levels than just `-1`, `0` and `1` in the future.


Artemis21:
CI test failures because prior to Python 3.8, argparse won't let you group short flags (`-v -v` -> `-vv`) when flag abbreviation is enabled.

Disabling flag abbreviation is a potentially breaking change, so I think we have to require users on older Python versions to type out `-v -v`.

Artemis21:
More test failures because apparently `echo 'Hello'` will echo `'Hello'` on Windows, whereas it is just `Hello` on UNIX.

Artemis21:
Passing at last 😅

nat-n (approved):
LGTM. Thanks for the fine work.

PR #36, opened 2021-09-11
Updated narg parsing. Documentation. Testing.

cigani:
- Modified nargs in the following ways:

1. nargs not explicitly provided are not passed to the generated kwargs dictionary
2. nargs can optionally have their type specified
3. types of nargs are checked against a simple lookup dictionary
4. tried to increase the feedback (error messages) where behavior might not be immediately transparent to the user 

- Updated the documentation to be a bit more clear about how default arguments are parsed
- Added a few more tests to ensure functionality of default and type specified arguments


tests/fixtures/dummy_project/pyproject.toml
```diff
@@ -18,11 +18,30 @@ cmd = "echo POE_ROOT:$POE_ROOT ${BEST_PASSWORD}, task_args:"
...
+    type = 'float'
+    help ='A float'
+
+
 [tool.poe.tasks]
 show_env = "env"
 greet = { script = "dummy_package:main" }
 greet-shouty = { script = "dummy_package:main(upper=True)" }
-count = { shell = "echo 1 && echo 2 && echo $(python -c 'print(1 + 2)')" }
+greet-keyed = { script ="dummy_package:args(greeting, user, optional, upper, default=default_value)", args = ["greeting", "user", "optional", "upper"]}
```
nat-n, line 41: What's default_value here?

poethepoet/task/script.py
```diff
@@ -102,10 +103,22 @@ def _resolve_args(cls, call_params: str, named_args: Dict[str, str]):
...
-        )
+                return {param: named_args.get(param, param)}
+
+        # I only want to filter out the empty strings that can come from the parser of named_args without having to modify named args
+        # test_script_task_omit_kwarg is an example of why this is necessary -- without removing the '' we have inconsistant behavior
+        # even if the arg isn't passed to poe it will be otherwise created here with an '' value
+        return {
+            k: v
+            for k, v in dict(
+                ChainMap(
```
nat-n, line 116: I think this could be simpler without ChainMap, am I missing something?
```suggestion
        return {
            key: value
            for param in call_params.split(",")
            for key, value in resolve_param(param.strip()).items()
            if value is not None
        }
```

Also single char variables (especially spanning multiple lines) are a pet peeve of mine.

nat-n (commented):
Hey thanks again for working on this. It's good to see progress!

I guess by nargs you mean "named args" rather than the argparse `nargs` option?

I've left a first round of comments. I still need to check it out and play around with it, maybe think through a few more scenarios for script args, and properly review the tests.

poethepoet/task/args.py
```diff
@@ -1,154 +1,186 @@
...
+arg_param_schema: Dict[str, Union[Type, Tuple[Type, ...]]] = {
+    "default": (str, int, float, bool),
+    "help": str,
+    "name": str,
+    "options": (list, tuple),
+    "required": bool,
+    "type": str,
+}
+
+arg_type_lookup: Dict[Any, Type] = {
```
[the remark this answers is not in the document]
cigani: I think there was at one point an issue I was having with mypy when I set it to str oddly enough. But In practice absolutely this would be a better solution. I'll switch it back and take a look at what was going on in mypy in more detail

README.rst
```diff
@@ -356,6 +356,40 @@ is also possible as demonstrated in the following example:
   [tool.poe.tasks]
   build = { script = "project.util:build(dest, build_version=version)", args = ["dest", "version"]
 
+For tasks that require specific types/defaults/help to be applied to arguments they can be constructed from subtables
+omission of variable type declaration will result in a string type.
+
+.. code-block:: toml 
+  
+  [tool.poe.tasks.A]
```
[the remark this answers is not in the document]
cigani: Yes this is a solid idea to make it a bit more 'humane'. I'll  patch this section up

poethepoet/task/script.py
cigani: I need to take a look. I can't recall exactly what issue I had when I wasn't using the ChainMap but it was an early bit of code I wrote so I might just not have given it sufficient thought. I'll look into this change.

cigani:
Thanks for the feedback nat. I'll try to realign the PR. Hopefully I'll get you something updated this weekend since I have to work this weekend anyway :D

nat-n (commented):
I think we shouldn't  require the use of kwargs (script.py:48), since it creates a coupling between the task definition and the function definition, whereby even if you define the call params using positional arguments, you actually have to make the variables names line up with the names from the function def anyway or it won't work. Also if you want to call a function that accepts `*args` you're screwed...

For context, a use-case I'd like to encourage is using poe to create a quick and convenient CLI for a python function from any random library with just a couple lines of config! But this means the user might not control the function signature they're interfacing with.

---

Also from seeing the default_value example in your test case and playing around I realise we have a bigger problem with how we handle call_params and need to rethink a bit. The way it works now (since my original implementation for named args) will cause a breaking change and loss of functionality vs 0.10.0.

That `default_value` in `script = "pkg:fun(foo=default_value)"` gets interpreted as a string is weird!

For consistency the following should work as one would naively expect:

```
[tool.poe.tasks.call]
script = "pkg:fun(foo='foo', bar=1, baz=1.0, qux=true, vroom=CLI_ARG, shell_path=environ.get('PATH'))"
args = ["CLI_ARG"]
```

Note this is assuming we put add `from os import environ` to the setup code which I think makes sense to enable this kind of usage, though that's probably a feature for after 0.11.0.

I think we should start with a failing test case like that one, and coordinate a bit so we can both push some changes to this branch without stepping on each others toes.

Thinking on solutions; One alternative approach would be to inject all the named argument values as locals in the scope where the function is called. Then maybe we don't need to edit the call_params string at all! Instead we just expect it to be valid python (maybe with a couple of checks against code injection shenanigans). The downside of this is that we need to think about how to deal with collisions. Eg. if a user wants a cli arg called locals. etc

Another option would be to inject the named args like is done for cmd tasks, e.g.

```
[tool.poe.tasks.call]
script = "pkg:fun(foo='foo', bar=1, baz=1.0, qux=true, vroom=${CLI_ARGS}, shell_path=environ.get('PATH'))"
args = ["CLI_ARG"]
```

Though this seems kind of weird...

Anything else probably requires some level of parsing python syntax to modify the call_args string with confidence; at least to handle primitive types, (which means quote matching etc, not just naively splitting on `,` as a first step 🙀  ). Maybe there is an easy way to achieve this... I think it's my preferred solution but more research is required.

What do you think?

Of course we can break this problem up, and leave parts of it to solve in follow up PRs/releases, as long as we don't box ourselves in.

tests/test_task_running.py
```diff
@@ -67,6 +67,91 @@ def test_script_task_with_hard_coded_args(
     assert result.stderr == ""
 
 
+def test_script_task_with_args(run_poe_subproc):
```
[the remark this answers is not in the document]
cigani: I'd thought of doing exactly that since the test file became a bit long in the tooth

tests/fixtures/dummy_project/pyproject.toml
cigani: > What's default_value here?

I think that's an interesting question that deserves a bit more detail because this is where behavior becomes a bit convoluted IMO.

`default=default_value`

a TLDR: This will be parsed as default=Str(default_value) and passed as such to the kwargs

---------------
A bit longer
When we pass this to script.py@102 we pass `call_params` and `named_params` the latter will have only the (4) arguments named in the square brackets of the TOML file the former will have all (5) arguments defined in the parentheses. So any value defined in those default values will by default be defined as a string object. The default=default_value will not be present in `argv` but it will be present in the `call_params` we actually pass off to the final python function script.py@54.

In summary for this bit the `argv` will only contain arguments defined the square brackets. This might allow a user to have some pre-defined behavior for a function with only some optional components. The downside of this is that any arguments not defined specifically in the TOML will be strings -- Thus any attempt at a Boolean flag will fail if a user anticipates that `flag=false` evaluates to `False`

cigani: > What's default_value here?

In this case default_value is "default_value" to poe. The default is sent to the `call_params` rather than `argv`

```        
        cmd = (
            "python",  # TODO: pre-locate python from the target env?
            "-c",
            "import sys; "
            "from importlib import import_module; "
            f"sys.argv = {argv!r}; sys.path.append('src');"
            f"import_module('{target_module}').{target_callable}({call_params or ''})",
        )
```



A simple example of why I could see my self using this pattern of behavior 
```
morning = { script = "dummy_package:main(greeting=Good Morning, name)" , args = ["name"] }
evening = { script = "dummy_package:main(greeting=Good Evening, name)" , args = ["name"] }
```

cigani:
>[tool.poe.tasks.call]
script = "pkg:fun(foo='foo', bar=1, baz=1.0, qux=true, vroom=${CLI_ARGS}, shell_path=environ.get('PATH'))"
args = ["CLI_ARG"]

As far as I can tell this doesn't work natively. All those "Default Values" will be interpreted as strings

cigani:
>[tool.poe.tasks.call]
script = "pkg:fun(foo='foo', bar=1, baz=1.0, qux=true, vroom=${CLI_ARGS}, shell_path=environ.get('PATH'))"
args = ["CLI_ARG"]

As far as I can tell this doesn't work natively. All those "Default Values" will be interpreted as strings

nat-n (approved):

Issue #27, opened 2021-05-10
Using bash instead of default shell


Algirdyz:
So right now poe checks the `$SHELL` variable and uses that shell to run the commands when using `shell` type.

Why was this the choice? Seems a bit strange. While I do use fish myself, I do not want to write poe tasks on my projects to also use it. I would like everyone to be able to jus run poe commands if they checkout the code without having fish.

`make` for example ignores `$SHELL`. And this makes sense because these command should be cross platform and not require a developer setup a specific environment to be able to use the project.

Anyway, this would be a breaking change so maybe configuration is the way to go? I could try make a pull request myself if there is some consensus on what's the best way to handle this.

nat-n:
Hi @Algirdyz, thanks for the feedback.

I think you're right. The use of `$SHELL` was an ill conceived attempt to ensure portability, and should probably be removed. Using bash as a lowest common denominator by default probably makes the most sense. 

I intend to do so as part of addressing #15. I'm OK with a breaking change for this (we're not 1.0.0 yet), though I'd want to make it smooth which implies mostly addressing #15 by supporting something like a task level option to specify a shell executable (or ordered list thereof to try). Unfortunately I think this a little complicated to get right.

I'm happy to accept help in the form of design feedback and PRs (though I'm likely to be quite picky about what gets merged from a PR).

ThatXliner:
Have you thought about using [shellingham](https://github.com/sarugaku/shellingham) or use a custom `POE_SHELL` variable?

castarco:
I wouldn't go for bash, but for `sh`, which is always a POSIX shell (bash might be not installed in some systems!)

nat-n:
@Algirdyz Thanks again for the feedback. 

A fix for this issue has just been released as [0.12.0](https://github.com/nat-n/poethepoet/releases/tag/v0.12.0)

Issue #15, opened 2020-11-25
Provide configuration options and saner defaults for shell kind to be used 


JosXa:
_Original issue title:_ Poe tries to enforce WSL on Windows 10

When running a poe `shell`-type task in Windows Powershell Core 7.0.3, it seems to attempt execution of the command inside the Windows Subsystem for Linux (WSL), but I am happy using the cross-platform Powershell for most use cases.

Using the following configuration:

```
[tool.poe.tasks]
test = { shell = 'test' }
```

It complains that the WSL is not installed:

```
PS C:\projects\someproject> poe test
Poe => test
Windows Subsystem for Linux has no installed distributions.
Distributions can be installed by visiting the Microsoft Store:
https://aka.ms/wslstore
```

nat-n:
I don't know much about windows dev environments, so I made a best guess about how to get POSIX shell scripts to work seamlessly on windows with a bit of trial and error, but not very exhaustive testing, and I may have missed something obvious.

The intention of the *shell* task type is POSIX shell scripts, so the canonical implementation is to pipe the content into a bash process. Therefore the strategy is to try find a bash executable to use by looking on the PATH, looking for git bash, or finally looking for WSL. It's all in this file: https://github.com/nat-n/poethepoet/blob/master/poethepoet/task/shell.py#L47

@JosXa What would you suggest doing differently?

JosXa:
Ah yes, that sounds perfect and actually looks like you've given it a lot of thought. I can also confirm that it works out of the box when WSL is installed. So then I'd argue we need an additional way to specify which shell we'd like to run on? The `cmd` also seems to be using the native Windows cmd.exe instead of powershell 🤔

JosXa:
So what I tried is to execute the following Windows command line:

```cmd
pwsh.exe -ExecutionPolicy Bypass -Command "cd .. && ls"
```

But when adding it to the poe tasks like so:

```toml
testps = {cmd = "pwsh.exe -ExecutionPolicy Bypass -Command \"cd .. && ls\""}
```

It'll just print out the command again, but not the result (setting ExecutionPolicy or not makes no difference):

```
Poe => pwsh.exe -Command "cd .. && ls"
cd .. && ls
```

nat-n:
> So then I'd argue we need an additional way to specify which shell we'd like to run on?

[...]

PR #45, opened 2021-12-21
Add ability to configure which type of shell to use for executing shell tasks

nat-n:
Addresses #15 and #27 

The new shell executable resolution logic only considers the $SHELL env var as a hint (if it appears to match the target shell type), and never explicitly looks for WSL on windows. See included updates to documentation for more details.

Example of a shell task using fish shell:

```toml
  [tool.poe.tasks.fibonacci]
  help = "Output the fibonacci sequence up to 89"
  interpreter = "fish"
  shell = """
    function fib --argument-names max n0 n1
      if test $max -ge $n0
        echo $n0
        fib $max $n1 (math $n0 + $n1)
      end
    end

    fib 89 1 1
  """
```

Note that there are a couple of significant limitations with the powershell integration that I don't know how to solve. Firstly when powershell starts it always prints some copyright boilerplate which is not really desirable for this use case, and secondly it doesn't propagate environment variables from the parent process which is quite a limitation (e.g. named args won't work). Suggestions are welcome from anyone more knowledgable about powershell.

### DIFF
--- a/poethepoet/task/shell.py
+++ b/poethepoet/task/shell.py
@@ -1,3 +1,4 @@
+from os import environ
 from shutil import which
 import sys
 from typing import (
@@ -71,7 +72,12 @@ class ShellTask(PoeTask):
     def _locate_interpreter(self, interpreter: str) -> Optional[str]:
         result = None
 
-        if interpreter == "posix":
+        # Try use $SHELL from the environment as a hint
+        shell_var = environ.get("SHELL", "")
+        if shell_var.endswith(f"/{interpreter}") and which(shell_var) == shell_var:
+            result = shell_var
+
+        elif interpreter == "posix":
             # look for any known posix shell
             result = (
                 self._locate_interpreter("sh")
--- a/poethepoet/task/shell.py
+++ b/poethepoet/task/shell.py
@@ -1,13 +1,26 @@
-import os
-import shutil
-import subprocess
-from typing import Dict, Mapping, Sequence, Tuple, Type, TYPE_CHECKING, Union
+from shutil import which
+import sys
+from typing import (
+    Any,
+    cast,
+    Dict,
+    Mapping,
+    Optional,
+    Sequence,
+    Tuple,
+    Type,
+    TYPE_CHECKING,
+    Union,
+)
 from ..exceptions import PoeException
 from .base import PoeTask
 
 if TYPE_CHECKING:
     from ..config import PoeConfig
     from ..context import RunContext
+
+KNOWN_INTERPRETERS = ("posix", "sh", "bash", "zsh", "fish", "pwsh", "python")
+DEFAULT_INTERPRETER = "posix"
 
 
 class ShellTask(PoeTask):
@@ -18,7 +31,7 @@ class ShellTask(PoeTask):
     content: str
 
     __key__ = "shell"
-    __options__: Dict[str, Union[Type, Tuple[Type, ...]]] = {}
+    __options__: Dict[str, Union[Type, Tuple[Type, ...]]] = {"interpreter": (str, list)}
 
     def _handle_run(
         self,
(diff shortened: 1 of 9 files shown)


poethepoet/task/shell.py
```diff
@@ -31,42 +41,104 @@ def _handle_run(
...
+            result = which("fish") or which("/bin/fish")
+
+        elif interpreter == "pwsh":
+            result = which("pwsh") or which("powershell")
+
+            # Specifically look in a known location on windows
+            if result is None and self._is_windows:
+                result = which(
+                    "C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell.EXE"
+                )
```
TBBle, line 111: I would suggest doing this only when using `powershell`, not `pwsh`, as the two are not fully cross-compatible.

TBBle:
> Firstly when powershell starts it always prints some copyright boilerplate which is not really desirable for this use case
Both `powershell` (Windows Powershell) and `pwsh` (PowerShell, aka PowerShell Core, the cross-platform one) support `-NoLogo` to hide the banner, if that's what you're looking for.

I'd also suggest distinguishing `pwsh` and `powershell`, as the former is cross-platform, the latter is Windows-only, and there's plenty of things that `pwsh` has that aren't in `powershell`, and also a few things that are in `powershell` but not `pwsh` (mostly Windows-specific stuff that calls through into .NET Framework, though). It might be reasonable to use `pwsh` when `powershell` is specified, but I definitely wouldn't user `powershell` when `pwsh` is specified.

> secondly it doesn't propagate environment variables from the parent process

This is bizarre, and not what I'd expect. I also can't replicate it trivially here, setting an env-var in cmd and then starting Powershell and Windows PowerShell see the env-var.

```cmd
Microsoft Windows [Version 10.0.19042.1415]
(c) Microsoft Corporation. All rights reserved.

C:\Users\paulh> echo %BOB%
%BOB%

C:\Users\paulh> set BOB="Hi"

C:\Users\paulh> pwsh -nologo

C:\Users\paulh
>echo $ENV:BOB
"Hi"
C:\Users\paulh
>exit

C:\Users\paulh> powershell -nologo
PS C:\Users\paulh> echo $ENV:BOB
"Hi"
PS C:\Users\paulh> exit

C:\Users\paulh>
```

So I don't think this is an endemic issue for PowerShell per-se.

I wonder if this is to do with the way it's launched? I don't see any command-line args, but perhaps it's different if you explicitly pass `-Command -` to execute stdin, in case that makes a difference?

TBBle:
I just grabbed this branch, and given
```
[tool.poe.tasks.echo_pwsh]
interpreter = "pwsh"
shell = "echo ${ENV:test_var}"
env = { test_var = "roflcopter" }
```

this test passes:
```
@pytest.mark.skipif(
    not (shutil.which("powershell") or shutil.which("pwsh")),
    reason="No powershell available",
)
def test_interpreter_pwsh(run_poe_subproc, is_windows):
    result = run_poe_subproc("echo_pwsh", project="shells")
    assert result.capture == ("Poe => echo ${ENV:test_var}\n")
    assert "roflcopter" in result.stdout
    # Noise due to my local setup, pwsh needs `-NoLogo -Command -` to suppress.
    #assert result.stderr == ""
```

Powershell doesn't AFAIK have an equivalent of `$0` (and it would probably fail on my system, since `pwsh` is installed from the .NET Global Tools, and so `(Get-Process -id $pid | Get-Item).Name` returns `dotnet.exe`.

And I reckon (but didn't test) that `-NoLogo -Command -` on the pwsh command-line would have correctly returned empty `result.stderr`.

This is testing on Windows 10, where a bunch of the other tests failed, I'm guessing your testing environment is Linux, so maybe PowerShell has a bug on Linux involving env-vars?

nat-n:
Hi @TBBle, thanks a lot for the input!

So it turns out I just didn't realise it was necessary to include the `$ENV:` instead of just `$` when referencing env vars 🙄.

Regarding powershell versions: I'm thinking that we should offer `{"pwsh7", "pwsh", "powershell"}`, where `"powershell"` matches any version that can be found with a preference for newer versions, and "pwsh" matches any version >=6, so if powershell 7 is present it will always be used. Does this sound sensible?

Do you agree that it's important to distinguish powershell 7 vs 6? I understand that 7 is backwards compatible since forever but  does introduce new syntax.

Is it reasonable to assume that most windows devs or CI environments would have just one version of powershell installed which would be >= 6?

Is it necessary to acknowledge the distinction between Powershell <= 5 and .NET Core editions other than the version number?

In my limited testing I've found that `which("bash")` is not an entirely reliable way of locating git bash. So I assume that it's worth having the typical locations of powershell hard coded in the same way. Could you help be determine the list of paths that are worth checking?

TBBle:
I think PowerShell 6 and 7 can be mixed together. The major distinction I've seen is between Windows PowerShell (versions < 6, based on .NET Framework, Windows only, `powershell.exe`), and PowerShell Core (versions 6+, based on .NET Core, cross platform, pwsh.exe, seems to just be officially referred to as PowerShell, "Core" has kind-of disappeared).

Basically, a bunch of new and nice stuff was added in PowerShell Core, but there are some things that may not work in PowerShell Core if they rely on .NET Framework APIs that are not present in .NET Core, so you can't (AFAIK) 100% assume a Windows PowerShell script works in PowerShell Core.

All interesting Windows machines (except Nanoserver container images, off-hand) have Windows PowerShell installed, it's been bundled for decades, at the hard-coded path already being tested. I think this is one of the reasons PowerShell Core is `pwsh.exe`, so you _know_ you're getting PowerShell Core, and `powershell.exe` always gives you Windows PowerShell.

I'm not sure if there's a really good default path for pwsh.exe, as there's a few different installation methods including portable .zip and per-user installs. I'd expect it to be in the path as pwsh.exe if it's installed though. And given how aggressively it prompts for upgrading, I wouldn't expect to hit many PowerShell 6 installs in the wild, particularly because it [fell out of support more than 12 months ago](https://docs.microsoft.com/en-us/powershell/scripting/install/powershell-support-lifecycle?view=powershell-7.2#powershell-end-of-support-dates).

That said, pwsh may not be that wide-spread. I wouldn't expect it on CI machines unless I specifically asked for (or bundled it in my CI container image, more likely).

In the end, I wouldn't fuss too much with the PowerShell Core versions, if `pwsh.exe` works, then it's PowerShell Core, and if it's somehow a support-expired version or similar, that's between the script author and the user.

poethepoet/task/shell.py
```diff
@@ -31,42 +42,120 @@ def _handle_run(
...
+
+            if result is None and interpreter == "powershell" and self._is_windows:
+                # Look for older versions of powershell
+                result = which("powershell") or which(
+                    environ.get("WINDIR", "C:\\Windows")
+                    + "\\System32\\WindowsPowerShell\\v1.0\\powershell.EXE"
+                )
+
+        elif interpreter == "python":
+            result = which("python") or which("python3") or sys.executable
```
TBBle, line 131: I'm curious under what circumstances one wouldn't use `sys.executable` when running a python script?

TBBle (approved):

poethepoet/task/shell.py
nat-n: I hope it's never needed, in that python or python3 should be on the path in any sane environment, but if somehow they're not then `sys.executable` gives us the one thing we know about for sure which is the python interpreter of the current process.

zmarffy: Forgive me if I am wrong, but I think @TBBle may have meant that using `sys.executable` is the most desirable option, rather than the last thing in the `or` chain.

nat-n: I see, maybe I misread. That is actually what I thought originally when implementing the script task type, but then [I changed my mind about it](https://github.com/nat-n/poethepoet/commit/fd85005c2062eac7863d00161aad408966fb8a0f#diff-6530823c1ef123e90f1bc9301d47e75c652a895a72e64f5f4e14f891e25c3f3aR26) for reasons that I can't recall with confidence.

However my understanding at this point is that given the many ways that python versions can be installed and managed on a system, it's a safer bet to use whatever is configured in the current environment rather than whatever the poe executable is using. 

For example poe might have been installed with the system python, but the task is being executed within a virtual env (perhaps activated by the PoetryExecutor or VirtualenvExecutor inside poe) which is probably more relevant and may reference a different python version (and site_packages) than the poe CLI is using.

TBBle: Ah, that's a good point. I've only used Poe installed via Poetry as a task runner, and did not consider a system-installed Poe.

PR #46, opened 2021-12-23
Add poetry plugin

nat-n:
Based off of #32 by @ThatXliner and refactored a bit.

- Uses the new command structure support in Poetry 1.2 to register commands with poe as a prefix. Also gives users the option to remove this (in which case they need to be careful not to have tasks that collide with poetry commands), or set it something else like `"run "` or `'x "` etc.
- Uses the virtualenv from poetry directly rather than calling poetry again in the task subprocess.
- Uses the default verbosity level from poetry unless it's set in pyproject.toml
- Includes a best effort hack to make open ended CLI options work with cleo via monkey patching
- Includes some enhancements for testing code

### DIFF
--- a/poethepoet/plugin.py
+++ b/poethepoet/plugin.py
@@ -1,33 +1,42 @@
 import shutil
 import subprocess
+from pathlib import Path
 from typing import Iterator
 
 import tomlkit
 from cleo.commands.command import Command
+from poetry.console import application as app
 from poetry.plugins.application_plugin import ApplicationPlugin
 
 from . import config
 
 
 def get_poe_task_names() -> Iterator[str]:
-    yield from tomlkit.loads(config.PoeConfig().find_pyproject_toml())["tool"]["poe"][
-        "tasks"
-    ].keys()
+    return (
+        name
+        for name in tomlkit.loads(
+            Path(config.PoeConfig().find_pyproject_toml()).read_text()
+        )["tool"]["poe"]["tasks"].keys()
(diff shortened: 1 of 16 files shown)


nat-n:
There's still an open question about how to handle passing of arguments to tasks. This doesn't work out of the box since tasks are registered as commands using cleo as a CLI library. I'm afraid that it may be necessary to refactor poethepoet to use Cleo throughout! Though this might still not help for passing arbitrary args to cmd tasks for instance. I haven't looked into this properly yet.

ThatXliner:
Ooh this looks promising

ThatXliner:
Looking good! Haven't tested it on my machine yet.

Looking from the code, I guess you got the plugin to invoke the `PoeThePoet` object.

poethepoet/plugin.py
```diff
@@ -0,0 +1,193 @@
+# pylint: disable=import-error
```
ThatXliner, line 1: I think we can remove this? This was to silence Pylint from the `try`/`except` hack I did before

nat-n: I added this to stop pylint complaining about cleo not being a direct dependency :)

nat-n:
@ThatXliner I think it's almost ready to go. If you could test it out a bit then that would be helpful :)

ThatXliner:
Screen Shot 2022-01-05 at 10 03 38 AM

just tested it, umm

Issue #51, opened 2022-02-07
"Script" tasks not working on windows?


tcamise-gpsw:
Hello. I am not able to get a "script" task to actually run on Windows (git bash). For sanity testing, I am trying to just run the "poe" task from this repo. It is configured in your `pyproject.toml` as:

```
  [tool.poe.tasks.poe]
  help   = "Execute poe from this repo (useful for testing)"
  script = "poethepoet:main
```

I am running it via:

```
poetry run poe poe
```

On windows, I get nothing:

```
[13:38:53] Tim@WINDOWS ~/temp/poethepoet/poethepoet (main)
$ poetry run poe poe
Poe => poe
```

On ubuntu, it works:

```
[13:38:26] tcamise@UBUNTU ~/temp/poethepoet (main)
$ poetry run poe poe
Poe => poe
Poe the Poet - A task runner that works well with poetry.
version 0.12.2

Result: No task specified.

USAGE
  poe [-h] [-v | -q] [--root PATH] [--ansi | --no-ansi] task [task arguments]

GLOBAL OPTIONS
  -h, --help     Show this help page and exit
  --version      Print the version and exit
  -v, --verbose  Increase command output (repeatable)
  -q, --quiet    Decrease command output (repeatable)
  -d, --dry-run  Print the task contents but don't actaully run it
  --root PATH    Specify where to find the pyproject.toml
  --ansi         Force enable ANSI output
  --no-ansi      Force disable ANSI output

CONFIGURED TASKS
  format         Run black on the code base
  clean          Remove generated files
  test           Run unit and feature tests
  test-quick     Run unit and feature tests, excluding slow ones
  types          Run the type checker
  lint           Run the linter
  style          Validate code style
  check-docs     Validate rst syntax in the docs
  check          Run all checks on the code base
  poe            Execute poe from this repo (useful for testing)
```

Am I doing something dumb here?

nat-n:
Hi @tcamise-gpsw, thanks for reporting this.

I don't seem to be able to reproduce this issue, using git bash on Windows 11 running in a VirtualBox.

I'm not sure what could be happening here, as long as "python" is on your path I'd expect it to work ([see relevant code](https://github.com/nat-n/poethepoet/blob/main/poethepoet/task/script.py#L51))

If you're able to debug it, or share anything potentially relevant about your environment then I'd be interested to see if there's something poethepoet can do to be more robust.

I'll leave this issue open for a while in case more details come to light.

[...]

PR #52, opened 2022-02-08
Remove newline when building command in script

tcamise-gpsw:
- Solves #51


nat-n:
This change should fix the issue in a more complete way: https://github.com/nat-n/poethepoet/commit/d35d66edc52fd3dc20f2a86d4ab177ef3efbe594

PR #57, opened 2022-02-27
Remove the upper bound on tomli 2.0

Parnassius:
Besides removing support for python 3.6, the only breaking changes in tomli 2.0 are in the first parameter of `load`/`loads`: it can no longer be passed by keyword, and in the case of `load` it must be a binary file object.  Poe already does that, so no further changes are required.


nat-n:
Thanks @Parnassius!

Issue #54, opened 2022-02-10
No way to map multiple args to single positional arg


wsascha:
Hey @nat-n, first of all thanks for that great tool! :heart: 

I couldn't find any functionality to map more than one arg to a single arg.

### Example

Maybe someone wants to have a python formatting task a la

```
[tool.poe.tasks.format]
cmd = "black ${files}" 
help = "Format files
args = [
  { name = "files", positional = true, default = ".", help = "List of files (optional)" },
]
```

Then, this task may be run e.g. via
- `poe format`
- `poe format "."`
- `poe format /path/to/file`
- `poe format "/path/to/file"`
- `poe format "/path/to/file/one /path/to/file/two"`, etc.

but not `poe format /path/to/file/one /path/to/file/two`. – Note the missing `"`'s in the example.

In my special case, I want to run tasks as pre-commit hooks using the `pre-commit` tool which, unfortunately, passes multiple filenames without the `"`'s.

Any idea how this can be resolved? – Maybe I have also just missed something :-)

nat-n:
Hi @wsascha, thanks for the feedback :)

I'm afraid this feature was on my list but didn't quit make it into the initial implementation of structured arguments. I'll take this as validation that it's worth adding, and so leave this issue open. Though there are a couple of higher priority features in progress so I'm not sure when I'll get around to it.

One thing that I'm not sure about though, is how multiple value arguments should be handled by tasks given that single argument values are expected to be strings which can be set as env vars or python variables within tasks. Only three solutions occur to me though none are very satisfactory:

1. take the list of arguments as a whitespace delimited string, in which case result is basically the same as your example of `poe format "/path/to/file/one /path/to/file/two"`, and makes it impossible to distinguish between multiple args and a single arg containing a space
2. store the argument with a variable for each item, e.g. `files0` and `files1`. This is kind of complicated, and requires you to know how many values you expect, in which case you might as well just define multiple positional args with those names.
3. Do both 1 & 2, this should cover more use cases but it's a bit complicated and inelegant, and maybe also pass the argument as a python list of script tasks. Probably the best solution though.

What do you think?

nat-n:
@wsascha Although looking again at your example, unless you're just using this simple example as a stand in for a more complicated use case... You don't need a named arguments at all. The following will work just fine:

```
[tool.poe.tasks.format]
cmd = "black" 
help = "Format files, optionally accepts any number of files to format"
```

wsascha:
[...]

PR #70, opened 2022-05-30
Add multiple option for args

nat-n:
WIP solution for #54 @wsascha

The new multiple option can be added to args to make them accept multiple values.

- only the last positional arg may have multiple set
- multiple option defaults to false
- if `multiple = true` and `required = false` then the result is `nargs = "*"`, or if `required = true` then `nargs = "+"`
- an integer value >= 2 may also be given to specify an exact number of values are required for an arg
- for script tasks the values are passed as a list, for command tasks they can be templated into the command (thus potentially passed as multiple arguments), shell tasks don't have any special support.
- All tasks also get the list of values as an env var in the form of a space delimited string

## Still to test:
- [x] created a task for testing in the scripts project, but still need to write thorough test cases yet in `test_script_tasks`
- [x] also need basic test cases for cmd, shell, and sequence tasks
- [x] what if multiple=true with type=boolean ?

## Also todo:
- [x] Fix other broken tests
- [x] Document new functionality in the readme
- [ ] Think of a nice way to indicate that an arg accepts multiple arguments when generating documentation for a task

### DIFF
--- a/poethepoet/task/args.py
+++ b/poethepoet/task/args.py
@@ -208,6 +208,11 @@ class PoeTaskArgs:
                 f"The multiple option for arg {arg_name!r} of {task_name!r}"
                 " must be given a boolean or integer >= 2"
             )
+        if multiple is not False and params.get("type") == "boolean":
+            return (
+                "Incompatible param 'multiple' for arg {arg_name!r} of {task_name!r} "
+                "with type: 'boolean'"
+            )
 
         return None
 
--- a/poethepoet/task/args.py
+++ b/poethepoet/task/args.py
@@ -26,6 +26,7 @@ arg_param_schema: Dict[str, Union[Type, Tuple[Type, ...]]] = {
     "positional": (bool, str),
     "required": bool,
     "type": str,
+    "multiple": (bool, int),
 }
 arg_types: Dict[str, Type] = {
     "string": str,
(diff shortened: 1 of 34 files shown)


poethepoet/task/args.py
```diff
@@ -26,6 +26,7 @@
     "positional": (bool, str),
     "required": bool,
     "type": str,
+    "multiple": (bool, int),
```
wsascha, line 29: Not sure if it's really necessary to have an int here. At least I don't know any use cases from tools I used.

tests/fixtures/scripts_project/pyproject.toml
```diff
@@ -99,6 +99,29 @@ greet = "pkg:greet"
     type = "boolean"
 
 
+  [tool.poe.tasks.multiple-value-args]
```
wsascha, line 102: Maybe we could also add a sequence task here.

wsascha (commented):
Hey, sorry for the late reply, I'm currently on vacation 😅  looks good to me over all. Thanks for your efforts, really appreciated! I left two minor comments below, let me know what you think.

poethepoet/task/args.py
nat-n: I'm not sure either. Though it seemed like a low cost feature to support, and imagined that there could be use-cases for  it.

PR #73, opened 2022-06-11
Change plugin add command in readme 

ThatXliner:
They changed how you add plugins


ThatXliner:
Oops, I missed a spot in https://github.com/nat-n/poethepoet/blob/main/README.rst#run-tasks-with-the-poe-cli

PR #77, opened 2022-07-01
Update Python constraints on README

danilofuchs:
Based on: https://github.com/nat-n/poethepoet/blob/main/pyproject.toml#L12 and https://github.com/nat-n/poethepoet/blob/main/.github/workflows/ci.yml#L41


nat-n:
Thanks for the fix! I don't know how I missed it :P

PR #81, opened 2022-07-17
Fix poetry plugin crash on windows when outside of project

### LINKED ISSUE #79: Plugin broke poetry
image

Steps to reproduce:
pip install poetry==1.2.0b2
poetry self add poetry-version-plugin
poetry self add poethepoet[poetry_plugin]

nat-n:
Fixes #79 

Also:
- Make the plugin handle unexpected errors more gracefully
- Update doc for installing plugin for correct shell syntax

### DIFF
--- a/poethepoet/plugin.py
+++ b/poethepoet/plugin.py
@@ -75,6 +75,23 @@ class PoeCommand(Command):
 class PoetryPlugin(ApplicationPlugin):
     def activate(self, application: Application) -> None:
         try:
+            return self._activate(application)
+        # pylint: disable=bare-except
+        except:
+            import os, sys
+
+            print(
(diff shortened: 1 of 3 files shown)


poethepoet/plugin.py
```diff
@@ -74,6 +74,23 @@ def get_poe(cls, application: Application, io: IO):
...
 class PoetryPlugin(ApplicationPlugin):
     def activate(self, application: Application) -> None:
+        try:
+            return self._activate(application)
+        # pylint: disable=bare-except
+        except:
+            import os, sys
+
+            print(
+                "error: poethepoet plugin failed to activate. Set DEBUG_POE_PLUGIN=1 for details.",
```
ThatXliner, line 84: I feel like this print should be in an else branch of the if statement

PR #92, opened 2022-09-23
Return cleo application exit code from monkeymatch

### LINKED ISSUE #91: When used as plugin poetry commands does not return proper exit code
When poethepoet is used as plugin poetry console commands stop returning proper exit codes. For example if poetry lock is not consistent with pyproject `poetry lock --check` should return exit code 1, instead it returns 0 when poethepoet is installed as plugin.

I think this issue is not same as #71, because my poethepoet tasks return proper exit codes anyway.

monim67:
The monkeypatch poethepoet plugin does on cleo application does not return the exit code returned by the command execution.

This PR returns the exit code from the application run patch function. Fixes #91.


nat-n:
Hi @monim67, thanks for the fix! 

Looks like I have some issues to fix with the CI pipeline, but Ill try get this into a release soon.

ThatXliner:
Errors seem unrelated

PR #98, opened 2022-10-12
docs: use poe --group dev instead of poe --dev

guilhermedeandrade:
`poe --dev` is deprecated


nat-n:
@guilhermedeandrade thanks for the fix!

PR #100, opened 2022-10-25
(🎁) only use tomli in python<3.11

KotlinIsland:


nat-n:
Hi @KotlinIsland,

Thanks for the contribution and sorry for the slow response. I intend to deal with this properly but don't have a lot of time lately. Having one less dependency for python 3.11 is a nice outcome.

However it looks like the change to the conftest is causing problems. Maybe we should still use tomli as a dev dependency to avoid that issue?

KotlinIsland:
@nat-n I'm not sure what's happening in the CI, but it looks like the `@pytest.mark.skipif(version_info < (3, 7)` is running on python 3.7

nat-n:
@KotlinIsland It looks like something is fishy the the CI environment. Rerunning the action on main:HEAD fails :( 

There's no reason to block this PR, so I've disable that test env for now, and will merge.

bluss:
For some reason this 'regressed' without comment in #102 and tomli is again a required dependency.

nat-n:
Thanks for highlighting this @bluss. It's been fixed in [0.29.0](https://pypi.org/project/poethepoet/0.29.0/) 🚀

bluss:
Oh, nice, thanks

Issue #43, opened 2021-12-17
Remove the upper bound on Python 4.0


pawamoy:
Setting this bound forces every project that depends on yours to also set it. Reasons why it's not obviously a good thing to do: https://iscinumpy.dev/post/bound-version-constraints/ (I'll let you form your own opinion of course).

nat-n:
Hi @pawamoy, I'm going to [plead innocent](https://github.com/nat-n/poethepoet/blob/main/pyproject.toml) on this one... unless I'm missing something?

pawamoy:
Poetry's `python = "^3.6.2"` is equivalent to `>=3.6.2,<4.0.0` :slightly_smiling_face:

nat-n:
Touché. And thanks for the link, it is a tricky topic and I thank you for the prompt to look into it.

It sounds like uncapping the python version might not be a bad idea (in principle).

However I just noticed that it's just as well we don't depend on ">=1.2.2" for tomli since [last week's release of 2.0.0](https://github.com/hukkin/tomli/blob/master/CHANGELOG.md#200) would have silently broken poe!

nat-n:
That said it looks like I've fallen into the trap of viral max python version constraints, due to using pylint with poetry:

```
  The current project's Python requirement (>=3.6.2) is not compatible with some of the required packages Python requirement:
    - pylint requires Python ~=3.6, so it will not be satisfied for Python >=4.0
```

So it looks like there's nothing to be done for now (and no urgency). I'd likely accept a PR to address this in future.

pawamoy:
Ah, I guess I'll open the same ticket at pylint then :joy: 
Alright, thanks :slightly_smiling_face:

pawamoy:
In fact, pylint 2.12.2 already removed the cap! https://github.com/PyCQA/pylint/blob/eec287fae66f8fc514d5daa9caee46fd0e0cb6d9/setup.cfg#L53
So if you update your pylint spec to remove the cap as well, that will be one less blocker :smile: (other deps might still have the cap)

nat-n:
You might have to keep going:

```
  SolverProblemError

  The current project's Python requirement (>=3.6.2) is not compatible with some of the required packages Python requirement:
    - astroid requires Python ~=3.6, so it will not be satisfied for Python >=4.0

  Because no versions of pylint match >2.12.2,<3.0.0
[...]

PR #101, opened 2022-11-04
(🎁) add python 3.11 to the CI and update 'dev-dependencies' to 'group.dev.dependencies'

KotlinIsland:


pyproject.toml
```diff
@@ -9,12 +9,12 @@ repository  = "https://github.com/nat-n/poethepoet"
 homepage    = "https://github.com/nat-n/poethepoet"
 
 [tool.poetry.dependencies]
-python = ">=3.7"
+python = "^3.7"
```
nat-n, line 12: We want to keep compatibility with python >=4 to avoid causing problems for dependent packages. See #43

nat-n (changes requested):

PR #105, opened 2022-11-18
Fix typo in --help output

howeaj:


nat-n:
That typo has been there for a long time and I never noticed! Thanks for the fix :)

PR #136, opened 2023-03-22
Release PR 0.19.0

nat-n:


tests/fixtures/scripts_project/pkg/__init__.py
```diff
@@ -59,3 +59,7 @@ class Scripts:
     class Deep:
         def fun(self):
             print("task!")
+
+
+def async_task(*args, **kwargs):
```
Gobot1234, line 64: Seems like theres a missing async here?

nat-n: yikes I committed the wrong version of the test! would you like to do the honors to fix it?

Gobot1234: I'd rather you did

Issue #119, opened 2023-01-17
Poe is not finding the `*.cmd` executables on Windows


ni-fgenois:
First of all, thank you for maintaining this really cool tool!

# Current behavior

On Windows, the following task fails to run as it can't find `npm.cmd`:
```toml
[tool.poe.tasks.my-node-command]
sequence = [
    {cmd="npm install"}
]
```

Specifying `{cmd="npm.cmd install"}` works fine, and `{cmd="node --help"}` also works as `node.exe` is in the path.

# Expected behavior

`poethepoet` should call `*.cmd` files as if they were regular executables.

# Workaround

Use a `switch` command

```toml
[tool.poe.tasks.my-node-command]
control.expr = "sys.platform"
[[tool.poe.tasks.my-node-command.switch]]
case = "win32"
sequence = [
    {cmd="npm.cmd install"},
]
[[tool.poe.tasks.my-node-command.switch]]
sequence = [
    {cmd="npm install"},
]
```

nat-n:
hi @ni-fgenois, thanks for the feedback and sorry for the delayed response.

My knowledge of windows is not strong, but I understand that normally the shell uses PATHEXT when searching the path for executables, but apparently the same thing doesn't work reliably when the cmd is run as a subprocess with Popen?

I guess the desirable behaviour would be if on windows it first tried to resolve the executable from the path using `shutil.where`? Would that work? Is that a good idea?

Happy to take advice on this one.

nat-n:
Hi @ni-fgenois,

I think I might have a fix for this in form of https://github.com/nat-n/poethepoet/commit/98d8a85dd1dccdc8054510037cc032e0899b6dc8, but I don't have a windows machine handy to test it.

Could you check if it works as expected now with the latest pre-release? https://pypi.org/project/poethepoet/0.18.2b2/

nat-n:
@ni-fgenois The fix was shipped in [0.19.0](https://pypi.org/project/poethepoet/0.19.0/) 🙂

ni-fgenois:
Really sorry for the late reply, I was off on sabbatical for a few months :S

I can confirm that @ameily 's fix solves my issue too, but unfortunately not the original fix.

@nat-n could you ship a new version with the aforementioned fix?

Thanks a lot for looking into this!

nat-n:
@ni-fgenois good timing, I just shipped @ameily 's fix included in [0.20.0](https://pypi.org/project/poethepoet/0.20.0/)

PR #142, opened 2023-04-23
win32: Resolve Executable to Absolute Path if not found in venv

ameily:
This PR addresses  #119 for `cmd`  tasks that use the poetry executor. I found that, on Windows, a task like the following wasn't working:

```toml
[tool.poe.tasks]
"check-spelling" = "npx cspell README.md CHANGELOG.md"
```

Running the task on Windows is failing because `npx` does not exist, it's actually `npx.cmd`:

```
poetry run poe check-spelling
# ..snip..
FileNotFoundError: [WinError 2] The system cannot find the file specified
```

The problem seems to be that the fix for #119 was done in [`PoeExecutor.execute`](https://github.com/nat-n/poethepoet/blob/main/poethepoet/executor/base.py#L135), however this isn't called by the poetry executor since it calls `PoeExecutor._execute_cmd()` directly, bypassing the fix. I've confirmed that this PR works for my case.

### DIFF
--- a/poethepoet/virtualenv.py
+++ b/poethepoet/virtualenv.py
@@ -1,4 +1,5 @@
 import os
+import shutil
 import sys
 from pathlib import Path
 from typing import Dict, Mapping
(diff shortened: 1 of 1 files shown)


nat-n (approved):
Hi @ameily, thanks the fix! I'll include it in a new release soon :)

Issue #16, opened 2021-01-04
Auto-globbing in cmd tasks makes it impossible to pass anything that looks like a glob to the called command


TBBle:
I'm having an issue with the following in my pyproject.toml
```
[tool.poe.tasks]
run-tests = "python -m unittest discover -v ./tests/"
```

This works fine:
```
PS D:\PS\services\projects\scaler> poetry run poe run-tests
Poe => python -m unittest discover -v ./tests/
.... <Tests run>
```

However, this fails:
```
PS D:\PS\services\projects\scaler> poetry run poe run-tests -p "test_unit*.py"
Poe => python -m unittest discover -v ./tests/ -p
usage: python.exe -m unittest discover [-h] [-v] [-q] [--locals] [-f] [-c] [-b] [-k TESTNAMEPATTERNS] [-s START] [-p PATTERN] [-t TOP]
python.exe -m unittest discover: error: argument -p/--pattern: expected one argument
```

Specifically, `"py_unit*.py"` has disappeared. This argument is a glob-pattern to be passed as-is to Pytest, so should not be expanded. The bash-like behaviour of "use glob as-is if not matched" would be less-surprising, producing
```
Poe => python -m unittest discover -v ./tests/ -p "py_unit*.py"
```

The behaviour of course is closer-to-expected if I happen to be in 'tests'
```
PS D:\PS\services\projects\scaler\tests> poetry run poe run-tests -p "test_unit*.py"
Poe => python -m unittest discover -v ./tests/ -p test_unit_all_the_tests.py
... <Runs tests from that file>
```

i.e. when I first set this up, it appeared to be working, but *only* because I happen to have just one test file matching the glob here. It also will then fail to match any subdirectory-held test files that aren't also named _test_unit_all_the_tests.py_.

In bash, `'py_unit*.py'` would mean "Don't expand this", but there's no way to communicate that to poethepoet as the quotes are already gone before it gets the argument.

```
[tool.poe.tasks]
run-unit-tests = "python -m unittest discover -v ./tests/ -p 'test_unit*.py'"
```
has the same problem, except it's always globbed from the top-level, not from the directory I happen to be in, leading to:
```
PS D:\PS\services\projects\scaler\tests> poetry run poe run-unit-tests -p "test_unit*.py"
Poe => python -m unittest discover -v ./tests/ -p -p test_unit_all_the_tests.py
```

where the *same glob* has expanded two different ways.

---

[...]

PR #145, opened 2023-05-04
cmd task: Strip quotes from token on Windows

ameily:
I found an issue related to #16 where, on Windows, a quoted token is being passed to the command still wrapped in quotes. This only happens on Windows because the `shelx.split(posix=False)` preserves them. You can verify this issue on Windows with the poethepoety `test-quick` task, which passes a quotes string to pytest and fails because the marker contains an invalid quote character:

```
Poe => pytest --cov=poethepoet -m "not slow"
# ... cut ...
ERROR: Wrong expression passed to '-m': "not slow": at column 1: unexpected character ""
```

I initially was seeing this with a call to cspell and have reproduced it with a small script / task.

```python
# args.py
import sys
print(sys.argv)
```

```toml
# pyproject.toml
[tool.poe.tasks.args]
cmd = "python args.py \"hello world\" \"some/glob/**/*.py\""
```

The results of running the `args` task show that, on Windows, the quotes are preserved when they should be stripped.

```
$ poetry poe args
Poe => python args.py "hello world" "some/glob/**/*.py"
['args.py', '"hello world"', '"some/glob/**/*.py"']
```

I tried using some of the examples from #16 but I couldn't get them to work in my case, either the glob was being expanded by poe or the quote characters were being passed to the subprocess.

### DIFF
--- a/poethepoet/task/cmd.py
+++ b/poethepoet/task/cmd.py
@@ -88,6 +88,11 @@ class CmdTask(PoeTask):
             if not is_quoted and _GLOBCHARS_PATTERN.match(cmd_token):
                 # looks like a glob path so resolve it
                 result.extend(glob(cmd_token, recursive=True))
+            elif is_quoted and self._is_windows:
+                # In this case, the cmd_token will still be wrapped in double or single
+                # quotes. We need to remove those otherwise they'll be pass into the
+                # command.
+                result.append(cmd_token[1:-1])
             else:
                 result.append(cmd_token)
         return result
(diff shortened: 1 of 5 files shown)


ameily:
It looks like the Windows tests are failing because they expect the quoted behavior in stdout. If you're ok with my change, I'll update the tests and get them passing.

nat-n:
Hi @ameily, thanks for working on this! If you fix the tests I'd be happy to merge it.

ameily:
This should be good to go. Thanks for maintaining poe and being responsive! I useit for several Windows and Linux projects.

nat-n (approved):

PR #148, opened 2023-05-12
fix: zsh completion without oh-my-zsh

jannismain:
I was going through the guide and stumbled upon this. Shouldn't it generate a completion script for `poe` instead of `poetry`?


nat-n (approved):
Oh man, that typo has been there for a while! thanks for the fix :)

Issue #158, opened 2023-07-26
Can't in-lined tasks from another file with cwd


roger-duthie-pivotal:
### Summary

If trying to use the syntax documented [here](https://poethepoet.natn.io/guides/include_guide.html#setting-a-working-directory-for-included-tasks):

```toml
[[tool.poe.include]]
path = "my_subproject/pyproject.toml"
cwd  = "my_subproject"
```

... there is an exception if any tasks are defined "in-line" like:

```toml
[tool.poe.tasks]
mytask = "mycooltask"
myothertask = "myothertask"
```

The issue is that these tasks are loaded as strings so that on line 291 `config.py` the `task_def` is a string and on line 305 there is a type error when `task_def["cwd"]` is attempted.

### Possible solution
To convert any string tasks to objects like: 

```python
for task_name, task_def in tasks.items():
    if isinstance(task_def, str):
        task_def = {
            "cmd": task_def
        }
        tasks[task_name] = task_def  # Not sure if this is necessary or if the dictionary will be updated...
    [...]
```

I can create a PR at some point if you think this is: 

- [ ] Real
- [ ] A good enough fix

nat-n:
Hi @rjaduthie, thanks for the issue! I'm out of the office the next week, but I'll look at this when I get back :)

roger-duthie-pivotal:
[like]  Roger Duthie reacted to your message:
________________________________

nat-n:
#159 is merged, this will be fixed in the next release.

PR #159, opened 2023-07-26
Convert included string-form tasks

rjaduthie:
Ensure that the tasks are in the dict-form before trying to assign a "cwd" to the definition.

This is related to issue #158 

Sorry no tests for it yet.  I'll come back and add those, if you think this fix is worthwhile.

### DIFF
--- a/poethepoet/config.py
+++ b/poethepoet/config.py
@@ -187,7 +187,7 @@ class PoeConfig:
         # Validate shell_interpreter type
         for interpreter in self.shell_interpreter:
             if interpreter not in self.KNOWN_SHELL_INTERPRETERS:
-                return (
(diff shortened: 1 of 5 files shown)


nat-n (approved):

nat-n:
Hi @rjaduthie, thanks for reporting this bug and proposing a fix. 

As you can see I amended your changes to add a failing test and a small refactor to enable a more comprehensive fix. The issue being that normally the default task type (for a task defined as a string) depends on the configured `default_task_type`, and I wanted to be consistent with this logic. Incidentally I just realised this detail is missing from the docs since I launched the new docs website 😮.

This fix will be included in the next release.

PR #161, opened 2023-08-03
Allow use of the original working directory for tasks

Spiffyk:
For some tasks, it may be useful to use the working directory from which the task has been executed, since they may be meant to operate on some files that the user points them to. This commit introduces a special '$exec_cwd' value for the 'cwd' task option, which does exactly that. Existing use-cases are unaffected by this addition and work as usual.

### DIFF
--- a/docs/tasks/options.rst
+++ b/docs/tasks/options.rst
@@ -19,7 +19,7 @@ The following options can be configured on your tasks and are not specific to an
   Provide one or more env files to be loaded before running this task.
 
 **cwd** :  ``str`` :ref:`📖<Running a task with a specific working directory>`
-  Specify the current working directory that this task should run with. The given path is resolved relative to the parent directory of the ``pyproject.toml``.
+  Specify the current working directory that this task should run with. The given path is resolved relative to the parent directory of the ``pyproject.toml``. May be set to a special value ``$exec_cwd`` to use the working directory from which the task has been executed.
 
 **deps** :  ``List[str]`` :doc:`📖<../guides/composition_guide>`
   A list of task invocations that will be executed before this one.
--- a/docs/tasks/options.rst
+++ b/docs/tasks/options.rst
@@ -19,7 +19,8 @@ The following options can be configured on your tasks and are not specific to an
   Provide one or more env files to be loaded before running this task.
 
 **cwd** :  ``str`` :ref:`📖<Running a task with a specific working directory>`
-  Specify the current working directory that this task should run with. The given path is resolved relative to the parent directory of the ``pyproject.toml``. May be set to a special value ``$exec_cwd`` to use the working directory from which the task has been executed.
+  Specify the current working directory that this task should run with. The given path is resolved relative to the parent directory of the ``pyproject.toml``, or it may be absolute.
+  Resolves environment variables in the format ``${VAR_NAME}``.
 
 **deps** :  ``List[str]`` :doc:`📖<../guides/composition_guide>`
   A list of task invocations that will be executed before this one.
@@ -104,7 +105,7 @@ In this case the referenced files will be loaded in the given order.
 Running a task with a specific working directory
 ------------------------------------------------
 
-By default tasks are run from the project root – that is the parent directory of the pyproject.toml file. However if a task needs to be run in another directory within the project then this can be accomplished by using the :toml:`cwd` option like so:
+By default tasks are run from the project root – that is the parent directory of the pyproject.toml file. However if a task needs to be run in another directory then this can be accomplished by using the :toml:`cwd` option like so:
 
 .. code-block:: toml
 
@@ -112,7 +113,17 @@ By default tasks are run from the project root – that is the parent directory 
     cmd = "npx ts-node -T ./build.ts"
     cwd = "./client"
 
-In this example, the npx executable is executed inside the :sh:`./client` subdirectory of the project, and will use the nodejs package.json configuration from that location and evaluate paths relative to that location.
+In this example, the npx executable is executed inside the :sh:`./client` subdirectory of the project (when ``cwd`` is a relative path, it gets resolved relatively to the project root), and will use the nodejs package.json configuration from that location and evaluate paths relative to that location.
+
+The ``cwd`` option accepts absolute paths and resolves environment variables in the format ``${VAR_NAME}``.
+
+Poe provides its own :sh:`$POE_PWD` variable that is by default set to the directory, from which poe was executed; this may be overridden by setting the variable to a different value beforehand. Using :sh:`$POE_PWD`, a task's working directory may be set to the one from which it was executed like so:
+
+.. code-block:: toml
+
+    [tool.poe.tasks.convert]
+    script = "my_project.conversion_tool:main"
+    cwd = "${POE_PWD}"
 
 
 Defining tasks that run via exec instead of a subprocess
--- a/docs/tasks/options.rst
+++ b/docs/tasks/options.rst
@@ -115,7 +115,7 @@ By default tasks are run from the project root – that is the parent directory 
 
 In this example, the npx executable is executed inside the :sh:`./client` subdirectory of the project (when ``cwd`` is a relative path, it gets resolved relatively to the project root), and will use the nodejs package.json configuration from that location and evaluate paths relative to that location.
 
-The ``cwd`` option accepts absolute paths and resolves environment variables in the format ``${VAR_NAME}``.
+The ``cwd`` option also accepts absolute paths and resolves environment variables in the format ``${VAR_NAME}``.
 
 Poe provides its own :sh:`$POE_PWD` variable that is by default set to the directory, from which poe was executed; this may be overridden by setting the variable to a different value beforehand. Using :sh:`$POE_PWD`, a task's working directory may be set to the one from which it was executed like so:
 
--- a/poethepoet/context.py
+++ b/poethepoet/context.py
@@ -1,3 +1,4 @@
+import os
 import re
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Dict, Mapping, Optional, Tuple
@@ -105,11 +106,15 @@ class RunContext:
     ) -> "PoeExecutor":
         from .executor import PoeExecutor
 
+        cwd_option = task_options.get("cwd", ".")
+        if cwd_option == '$exec_cwd':
+            cwd_option = os.getcwd()
+
         return PoeExecutor.get(
             invocation=invocation,
             context=self,
             env=env,
-            working_dir=self.project_dir / task_options.get("cwd", "."),
+            working_dir=self.project_dir / cwd_option,
             dry=self.dry,
             executor_config=task_options.get("executor"),
             capture_stdout=task_options.get("capture_stdout", False),
--- a/poethepoet/context.py
+++ b/poethepoet/context.py
@@ -1,4 +1,3 @@
-import os
 import re
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Dict, Mapping, Optional, Tuple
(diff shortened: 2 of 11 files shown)


poethepoet/context.py
```diff
@@ -105,11 +106,15 @@ def get_executor(
...
+        cwd_option = task_options.get("cwd", ".")
+        if cwd_option == '$exec_cwd':
+            cwd_option = os.getcwd()
+
         return PoeExecutor.get(
             invocation=invocation,
             context=self,
             env=env,
-            working_dir=self.project_dir / task_options.get("cwd", "."),
+            working_dir=self.project_dir / cwd_option,
```
nat-n, line 117: The logical flow here isn't quite right; I if `cwd_option` is set from `os.getcwd()`  then it doesn't make sense to join it with `self.project_dir`.

poethepoet/context.py
```diff
@@ -105,11 +106,15 @@ def get_executor(
     ) -> "PoeExecutor":
         from .executor import PoeExecutor
 
+        cwd_option = task_options.get("cwd", ".")
+        if cwd_option == '$exec_cwd':
```
nat-n, line 110: Stylistically having the magic value look like an env var is inconsistent, and could be confusing. 

So I would prefer one of either:
A) making the cwd option accept an environment variable (the same way [some other options do](https://github.com/nat-n/poethepoet/blob/75ef547e8c1802e96661fd9b5c2da8d20930a5ec/poethepoet/task/args.py#L269)), which can be passed `"${PWD}"` as a value (with some special logic to make it work the same seamlessly on windows); or
B) use a magic keyword that doesn't look like an env var like `"parent_cwd"` (I'm undecided on the best naming semantics for this because normally just "cwd" would mean what we want here :/ )

At this moment I'm not convinced any particular solution is best... more thought required. One nice thing about option A is that it could make it possible for the cwd to depend on a user defined argument.

docs/tasks/options.rst
```diff
@@ -19,7 +19,7 @@ The following options can be configured on your tasks and are not specific to an
   Provide one or more env files to be loaded before running this task.
 
 **cwd** :  ``str`` :ref:`📖<Running a task with a specific working directory>`
-  Specify the current working directory that this task should run with. The given path is resolved relative to the parent directory of the ``pyproject.toml``.
+  Specify the current working directory that this task should run with. The given path is resolved relative to the parent directory of the ``pyproject.toml``. May be set to a special value ``$exec_cwd`` to use the working directory from which the task has been executed.
```
nat-n, line 22: It's more important that the linked section of the docs includes this detail rather than just this stub.

nat-n (changes requested):
Hi @Spiffyk, thanks for the PR, I think this feature is a good idea!

I left some comments about things that need to be improved before this can be merged. We also need to cover this feature with a test, and the CI is failing on coding standards (`poe format` should fix it).

If you're keen to own this then great (and please let me know), otherwise I should have some time soon to fix it up before merging.

Spiffyk:
Hi, @nat-n, I'll be happy to own this and fix it up. Thanks for the review, I'll look into all of it soon!

poethepoet/context.py
Spiffyk: Functionally, as [documented in `pathlib`](https://docs.python.org/3/library/pathlib.html#operators), this should be fine - if the right operand is an absolute path, the left operand gets discarded. The result of `os.getcwd()` will surely always be an absolute path, at least in my mind it wouldn't make much sense if it weren't - although the docs don't explicitly say that.

That said, I can rewrite it to be a little more explicit, if you wish.

poethepoet/context.py
Spiffyk: I don't hate option A, although we would need to ensure that the value of `PWD` is the actual startup working directory. `PWD` is normally defined by the shell, which may not always be present, I presume. Maybe we should use a different name (like `PARENT_PWD` as you suggested for option B) so that it does not confuse users, and is explicitly the type of value we want it to be, since it would be defined internally.

As for option B, I would still prefer to have some special character that is unlikely to be a part of an actual directory name (however unlikely it may be that a directory will be called `cwd`). Maybe something like `##parent_cwd##` would work?

nat-n: Thinking about this again, I'm leaning more towards option A. I agree that the issue of having to make the magic value not be path like is a bit of a design smell, but I also like the incidental power of accepting env vars :). 

As for PWD always being available from the environment... maybe the most elegant/delightful solution would be to make `${PWD}` always resolve to the result of `os.getcwd()`, whether or not it's inside a shell that works that way. What do you think?

Probably this would imply extending the `EnvManager::fill_template` function to accept an optional kwarg that could accept a mapping of override variables like `{"PWD": os.getcwd()}`. Either that or we add a new env var –maybe `$POE_INVOKE` to complement the existing `$POE_ROOT` that points to the dir with the pyproject.toml.

poethepoet/context.py
nat-n: Understood, it's probably work fine as it is... still could be clearer/more correct.

poethepoet/context.py
Spiffyk: Yea, thinking about it more, I'm also liking option A quite a bit - could lead to some interesting use-cases.

As for `PWD`, I'd personally still lean more towards defining a custom variable, maybe `${POE_PWD}`? This could work the following way: if it's not provided (i.e. the user did not set it beforehand), we use `os.getcwd()` to define it; if it *is* provided, we may well reuse the provided value. This way the semantics would be pretty clear when suitably documented. I feel like the shell's `$PWD` variable is kind of dubious as it is and I'd be more inclined to avoid using it if we can.

For what it's worth, I feel like extending the `EnvManager` method would be slightly overcomplicating the whole matter, more so than it needs to be. But I don't mean to impose, if you think it would be useful perhaps in more places, we may do it that way :)

nat-n: OK, lets go with setting POE_PWD as you suggest in EnvManager [along side POE_ROOT](https://github.com/nat-n/poethepoet/blob/main/poethepoet/env/manager.py#L52). As to what happens if POE_ROOT is already set: we need to think through the implications of always setting it to `os.getcwd()` (or not) in scenarios where the poe CLI is invoked by another poe task (e.g. monorepo usecase). I'm not sure it makes sense to consider that the user might explicitly set POE_PWD, but inheriting it from the parent poe task *might* make sense.

poethepoet/context.py
Spiffyk: Made this more explicit, how does it look to you?

poethepoet/context.py
Spiffyk: I added POE_PWD as instructed, and covered it by tests as much as I could tell would be useful. As far as I can tell, it seems to all be working, but please do let me know if I have overlooked something :)

docs/tasks/options.rst
Spiffyk: Updated the docs. Let me know what you think :)

nat-n (changes requested):
Nice work, it's almost there :)

tests/test_includes.py
```diff
@@ -88,7 +91,8 @@ def test_monorepo_contains_only_expected_tasks(run_poe_subproc, projects):
         "CONFIGURED TASKS\n"
         "  get_cwd_0      \n"
         "  get_cwd_1      \n"
-        "  get_cwd_2      \n\n\n"
+        "  get_cwd_2      \n"
```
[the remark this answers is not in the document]
Spiffyk: `main` currently seems to be up to date with `development`, so I just retargeted the PR.

tests/fixtures/cwd_project/pyproject.toml
```diff
@@ -1,3 +1,11 @@
 [tool.poe.tasks.cwd]
 cmd = "poe_test_pwd"
 cwd = "./subdir/foo"
+
+[tool.poe.tasks.cwd_env]
+cmd = "poe_test_pwd"
+cwd = "./subdir/${BAR_ENV}"
```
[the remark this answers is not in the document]
Spiffyk: Good catch, it helped find a bug, as I was using `self.env` in `get_executor` instead of the `env` parameter. For reasons I did not care too much to track down `self.env` does not have the variables generated from program arguments, but the `env` parameter does.

Anyway, I replaced `self.env` with `env` and it seems to be working correctly now.

poethepoet/env/manager.py
```diff
@@ -51,6 +52,9 @@ def __init__(
 
         self._vars["POE_ROOT"] = str(self._config.project_dir)
 
+        if "POE_PWD" not in self._vars:
+            self._vars["POE_PWD"] = os.getcwd()
```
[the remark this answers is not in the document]
Spiffyk: So I did this, it broke a few tests, so I (hopefully correctly) adapted them, but there is one in `test_includes.py` that I'm not sure how to fix properly.

It seems that `run_poe_subproc()` always searches for tasks in the current working directory, but these tests should ideally search for them in a separate directory, as poe already allows us to do with the `--root` option. However, I'm not sure how to pass that down to the command. Would you mind taking a look?

tests/test_includes.py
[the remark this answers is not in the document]
Spiffyk: Oops, my bad, I fetched the wrong remote. Rebase coming right up

docs/tasks/options.rst
```diff
@@ -104,15 +105,25 @@ In this case the referenced files will be loaded in the given order.
...
 .. code-block:: toml
 
     [tool.poe.tasks.build-client]
     cmd = "npx ts-node -T ./build.ts"
     cwd = "./client"
 
-In this example, the npx executable is executed inside the :sh:`./client` subdirectory of the project, and will use the nodejs package.json configuration from that location and evaluate paths relative to that location.
+In this example, the npx executable is executed inside the :sh:`./client` subdirectory of the project (when ``cwd`` is a relative path, it gets resolved relatively to the project root), and will use the nodejs package.json configuration from that location and evaluate paths relative to that location.
+
+The ``cwd`` option accepts absolute paths and resolves environment variables in the format ``${VAR_NAME}``.
```
nat-n, line 118:
```suggestion
The ``cwd`` option also accepts absolute paths and resolves environment variables in the format ``${VAR_NAME}``.
```

tests/test_includes.py
```diff
@@ -153,3 +157,11 @@ def test_monorepo_runs_each_task_with_expected_cwd(
     else:
         assert result.stdout.endswith("/tests/fixtures/monorepo_project/subproject_2\n")
     assert result.stderr == ""
+
+    result = run_poe_subproc("get_cwd_3", project="monorepo", cwd=projects["example"])
+    assert result.capture == "Poe => import os; print(os.getcwd())\n"
+    if is_windows:
+        assert result.stdout.endswith("poethepoet\\tests\\fixtures\\example_project\n")
```
nat-n, line 164:
```suggestion
        assert result.stdout.endswith("\\tests\\fixtures\\example_project\n")
```

tests/test_includes.py
```diff
@@ -153,3 +157,11 @@ def test_monorepo_runs_each_task_with_expected_cwd(
...
+    else:
+        assert result.stdout.endswith("poethepoet/tests/fixtures/example_project\n")
```
nat-n, line 166: I recently changed the examples in this test because it makes the test dependent on the name of the checkout directory.

```suggestion
        assert result.stdout.endswith("/tests/fixtures/example_project\n")
```

tests/test_includes.py
```diff
@@ -153,3 +157,11 @@ def test_monorepo_runs_each_task_with_expected_cwd(
     else:
         assert result.stdout.endswith("/tests/fixtures/monorepo_project/subproject_2\n")
     assert result.stderr == ""
+
+    result = run_poe_subproc("get_cwd_3", project="monorepo", cwd=projects["example"])
```
nat-n, line 161: You can pass `--root` to make this test pass like so:
```suggestion
    result = run_poe_subproc(
        "--root", 
        str(projects["monorepo/subproject_3"]), 
        "get_cwd_3", 
        cwd=projects["example"]
    )
```

nat-n (changes requested):

nat-n (changes requested):
Thanks for the fast turn around.  Just one little issue, then I think it'll be good to go.

poethepoet/env/manager.py
```diff
@@ -51,6 +53,10 @@ def __init__(
 
         self._vars["POE_ROOT"] = str(self._config.project_dir)
 
+        if "POE_PWD" not in self._vars:
+            self._vars["POE_PWD"] = str(cwd or os.getcwd())
+        self.cwd = self._vars["POE_PWD"]
```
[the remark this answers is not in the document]
Spiffyk: Right, that makes sense. Thanks!

nat-n (approved):
Thanks for contributing this feature @Spiffyk 🙂

It’ll be included in the next release.

Spiffyk:
Thanks for the help to get this in properly! 🙂

PR #163, opened 2023-08-24
Preserve '--' pseudo argument when passing to tasks

Spiffyk:
This commit makes it so that when a user passes the `--` pseudo-argument to a task, it is preserved for further processing by the task itself. Various commands may use `--` to e.g. pass arguments to another subprogram, or (e.g. in the case of `argparse`) to differentiate positional arguments from options.
    
Note that this change potentially breaks compatibility, as some already implemented tasks may rely on the `--` pseudo-argument being dropped, in fact, one of the tests was written that way. However, I strongly believe this is actually a bug and I decided to update the test as well.

### DIFF
--- a/poethepoet/task/cmd.py
+++ b/poethepoet/task/cmd.py
@@ -39,7 +39,7 @@ class CmdTask(PoeTask):
             # dash token: `--`
             try:
                 split_index = extra_args.index("--")
-                extra_args = extra_args[split_index + 1 :]
+                extra_args = extra_args[split_index:]
             except ValueError:
                 extra_args = tuple()
 
--- a/poethepoet/ui.py
+++ b/poethepoet/ui.py
@@ -133,8 +133,23 @@ class PoeUi:
         return parser
 
     def parse_args(self, cli_args: Sequence[str]):
+        pre_pa_args: Sequence[str]
+        post_pa_args: Optional[Sequence[str]]
+        try:
+            pa_index = cli_args.index("--")
+            pre_pa_args = cli_args[:pa_index]
+            post_pa_args = cli_args[pa_index:]
+        except ValueError:
+            pre_pa_args = cli_args
+            post_pa_args = None
+
         self.parser = self.build_parser()
-        self.args = self.parser.parse_args(cli_args)
+        self.args = self.parser.parse_args(pre_pa_args)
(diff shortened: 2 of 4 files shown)


Spiffyk:
Oh, only now do I realize that Poetry itself also throws the `--` pseudo-argument away, so that's quite a bummer, and renders this almost useless :(

Spiffyk:
Alright, so it's a known issue on `poetry`'s side and has a straightforward workaround using `poetry run -- [whatever]`. Would there be interest in having this behaviour adjusten on `poethepoet`'s side then?

nat-n:
Hi @Spiffyk, thanks for the PR.

I'm not sure I really understand the issue you're encountering. I take it your using poe via the poetry plugin and that's where you're getting behaviour different that what you expect, but this is due primarily to how poetry handles arguments?

Could you provide an example of a task to reproduce the issue?

Here's how argument parsing it intended to work today wrt `--`.

1. For tasks _without_ `args` declared it is treated like any other argument, .e.g.
```toml
[tool.poe.tasks.asdf]
cmd = "echo"
```
```sh
$ poe asdf hello -- goodbye
Poe => echo hello -- goodbye
hello -- goodbye
$ poetry poe asdf hello -- goodbye
Poe => echo hello -- goodbye
hello -- goodbye
```

2. For tasks _with_ `args` declared it is parsed to indicate that the remaining arguments should be append to to the end of the command (for a cmd task specifically), (documented [here](https://poethepoet.natn.io/guides/args_guide.html#passing-free-arguments-in-addition-to-named-arguments), covered by the test you modified) e.g.

```toml
[tool.poe.tasks.asdf]
cmd = "echo $foo and "
args = ["foo"]
```
```sh
$ poe asdf --foo hello -- goodbye
Poe => echo hello and goodbye
hello and goodbye
$ poetry poe asdf --foo hello -- goodbye
Poe => echo hello and goodbye
hello and goodbye
```

Does this make sense? Can you demonstrate how this doesn't work with your use case?

Spiffyk:
Oh, I see. The issue was with poetry all along and I somehow thought the same problem was present in poethepoet. I re-tested it locally with the `poetry run -- [cmd ...]` workaround and it seems to work.

Closing the PR as it is invalid.

PR #172, opened 2023-10-16
Fix document of switch tasks

kzrnm:
`control.expr` is defined in the `build` task and `switch` is defined in `platform_dependent` task, so there is a discrepancy in definitions.

I fixed this by replacing `platform_dependent` with `build`.


nat-n (approved):
@kzrnm Thanks for fixing my copy paste errors!

PR #173, opened 2023-10-16
📝 add information about referencing vars from PowerShell to docs

mason3k:
It is not totally obvious how to reference variables passed to a shell that uses a PowerShell interpreter. The trick is you need to access them as environment variables. I added a brief blurb on how to do this to the docs. 

I rebuilt the docs and reviewed my change visually:
image

The tip at the bottom is what I added.


nat-n:
Hey @mason3k, thanks for contributing 😄

PR #175, opened 2023-10-17
Skip PoetryExecutor if poetry is not found

kzrnm:
poe cannot run `PoetryExecutor` if poetry is set as an alias instead of `$PATH`.

I want to skip `PoetryExecutor` when poetry is not found.

### DIFF
--- a/poethepoet/executor/base.py
+++ b/poethepoet/executor/base.py
@@ -70,6 +70,10 @@ class PoeExecutor(metaclass=MetaPoeExecutor):
         self._is_windows = sys.platform == "win32"
 
     @classmethod
+    def fits_in(cls, context: "RunContext") -> bool:
+        return True
+
+    @classmethod
     def get(
         cls,
         invocation: Tuple[str, ...],
@@ -96,7 +100,6 @@ class PoeExecutor(metaclass=MetaPoeExecutor):
         by making some reasonable assumptions based on visible features of the
         environment
         """
-        from ..virtualenv import Virtualenv
 
         config_executor_type = context.executor_type
         if executor_config:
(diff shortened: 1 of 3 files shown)


nat-n (commented):
Hi @kzrnm , I'm curious to know more about your use case for this? Though I guess this behaviour makes sense if you have a project with a poetry managed local venv, but just don't have poetry somewhere discoverable for whatever reason.

Though I think falling back to the simple executor could be surprising.

kzrnm:
```powershell
# Microsoft.PowerShell_profile.ps1
Set-Alias poetry "${env:APPDATA}\pypoetry\venv\Scripts\poetry.exe"
```

I use poetry via alias on PowerShell, so that python interpreter cannot find poetry command and raise `FileNotFoundError` at PoetryExecutor.

https://github.com/nat-n/poethepoet/blob/903ed9af0dd313a6da32f24ec4357101435976d8/poethepoet/executor/poetry.py#L80-L85

## SimpleExecutor work properly.

In `poetry run`, poetry adds its venv directory to the `$PATH` of the subprocess.

### Example

```toml
[tool.poe.tasks.env]
shell = "/usr/bin/env | /usr/bin/grep ^PATH="
```

```
# Git bash
$ env PATH="" $APPDATA/pypoetry/venv/Scripts/poetry run poe env
Poe => /usr/bin/env | /usr/bin/grep ^PATH=
PATH=/mingw64/bin:/usr/bin:/c/Users/kzrnm/bin:/z/poethepoet/.venv/Scripts
```

nat-n:
Thanks for the update. I'm happy to merge this once we work out what's going on with the tests.

Although maybe it could still be improved:

I wonder if it would be a good idea to log a warning when poe fallback to the SimpleExecutor in a poetry governed project because the poetry executable can't be found.

Alternatively I wonder if it would be better still if poe were to specifically check for poetry at `$APPDATA/pypoetry/venv/Scripts/poetry` if running on windows and poetry doesn't show up on the path?

nat-n (approved):

nat-n (approved):
Thanks for the update. I'm happy to merge this once we work out what's going on with the tests.

Although maybe it could still be improved:

I wonder if it would be a good idea to log a warning when poe fallback to the SimpleExecutor in a poetry governed project because the poetry executable can't be found.

Alternatively I wonder if it would be better still if poe were to specifically check for poetry at $APPDATA/pypoetry/venv/Scripts/poetry if running on windows and poetry doesn't show up on the path?

kzrnm:
@nat-n I updated `fix/poetry` to the latest. 

i don't know why, but CI have worked.

PR #205, opened 2024-03-15
Fixes Windows error message due to incorrect type comparison.

BertrandBordage:
On Windows, I only see `Some dependencies may be missing from your system.` even though the value of `config_value` is `('posix',)`. This is because the type comparison is wrong, it checks that the tuple `('posix',)` is in `('posix', 'bash')`, instead of checking the string `posix`.

This PR fixes it by doing a set subset check.


nat-n (approved):

nat-n:
Hey @BertrandBordage, thanks for the contribution!

PR #207, opened 2024-03-26
Fix docs plugin link

worldworm:
Hi there,
I have noticed that the following URL is broken on this page: https://poethepoet.natn.io/poetry_plugin.html
https://python-poetry.org/docs/master/plugins/#using-plugins

I have changed it to:
https://python-poetry.org/docs/main/plugins/#using-plugins

Have a nice day 😃 
Great tool btw!


nat-n (approved):

nat-n:
Thanks for the fix @worldworm

PR #211, opened 2024-04-05
Fix plugin get directory from poetry

worldworm:
https://github.com/nat-n/poethepoet/discussions/209

With a bit of tinkering, I found `application.poetry.pyproject_path`. This contains the path to pyproject.toml

Using the example of #209:
Running `poetry --directory backend install` from the dir `project/` the attribute `application.poetry.pyproject_path` has the value `backend/pyproject.toml`
Running `poetry install` from the dir `project/backend/` the attribute `application.poetry.pyproject_path` has the value `/project/backend/pyproject.toml`

I hope I haven't missed anything essential 😃

### DIFF
--- a/poethepoet/plugin.py
+++ b/poethepoet/plugin.py
@@ -56,7 +56,11 @@ class PoeCommand(Command):
         except:  # noqa: E722
             poetry_env_path = None
 
-        cwd = Path().resolve()
+        # Get the cwd from poetry to ensure '--directory subdir' usage
+        try:
+            cwd = application.poetry.pyproject_path
+        except AttributeError:
+            cwd = Path().resolve()
         config = PoeConfig(cwd=cwd)
 
         if io.output.is_quiet():


nat-n:
Hey @worldworm, thanks for contributing this fix!

It might be a good idea to add a test case for this to make sure it stays fixed: https://github.com/nat-n/poethepoet/blob/main/tests/test_poetry_plugin.py What do you think?

worldworm:
Oh, that's an excellent idea. I totally forgot about those. 😄 

Unfortunately, after several attempts, I still haven't managed to get the test cases green. The `--directory` command doesn't seem to be recognized.
I have a feeling that it might have something to do with [the older version of poetry that is used for the test cases](https://github.com/nat-n/poethepoet/blob/main/tests/fixtures/packages/poetry-1.2.1-py3-none-any.whl).
https://github.com/nat-n/poethepoet/blob/7fecb34a0479bfe4250971e6524b22b9971fd6e5/tests/conftest.py#L257

I'll take a closer look tomorrow and get back to you.

nat-n:
> 
> https://github.com/nat-n/poethepoet/blob/7fecb34a0479bfe4250971e6524b22b9971fd6e5/tests/conftest.py#L257
> 

Indeed, hopefully it's easy to update the poetry version for testing!

worldworm:
So, it seems to me that the attribute from which I organised the path (`application.poetry.pyproject_path`) was introduced in a newer poetry version that is > 1.2.1. That explains why i had no success with test cases yesterday.

I have now simply downloaded a new poetry version ([the wheel directly from here](https://github.com/python-poetry/poetry/releases/tag/1.8.2)). If I understand correctly, this is only used for the test cases and should therefore have no effect on other things.

However, this shows that my changes are not breaking changes, as in the case of an older poetry version my except statement applies and the cwd is calculated as before. 😃

nat-n (approved):
LGTM. Thanks for the contribution!

PR #225, opened 2024-06-15
Fix handling comments in multiline cmds

### LINKED ISSUE #224: Unable to define a command which spans multiple lines
Hi, I noticed the [documentation mentions](https://poethepoet.natn.io/tasks/task_types/cmd.html#glob-expansion) that it's possible to define a command across multiple lines with comments and thus I was to configure the following command

```toml
[tool.poe.tasks.test-with-coverage]
help = "Run tests and record coverage"
cmd = """
pytest
    --cov=package                           # record coverage in package
    --cov-report=xml:.reports/coverage.xml  # save xml for coverage upload
    --cov-report=html:.reports/html         # save html files for local dev use
    --cov-branch                            # record coverage across logical branches
    --cov-context=test                      # show which tests cover specific lines in the code (available in HTML)
"""
```

However `poe` does not seem to be happy about it and throws me the following error:

```
Error: Invalid cmd task 'test-with-coverage' includes multiple command lines
```

---

Then, I tried configuring the example given in the docs

```toml
[tool.poe.tasks]
clean = """
rm -rf ./**/*.pyc
       ./**/__pycache__    # this will match all __pycache__ dirs in the project
"""
```

However this resulted in an exception

```
$ poe clean
  File "/media/pipx-3.8/bin/poe", line 8, in <module>
    sys.exit(main())
  File "/media/pipx-3.8/venvs/poethepoet/lib/python3.8/site-packages/poethepoet/__init__.py", line 35, in main
    result = app(cli_args=sys.argv[1:])
  File "/media/pipx-3.8/venvs/poethepoet/lib/python3.8/site-packages/poethepoet/app.py", line 126, in __call__
    return self.run_task(task) or 0
  File "/media/pipx-3.8/venvs/poethepoet/lib/python3.8/site-packages/poethepoet/app.py", line 174, in run_task
    return task.run(context=context)
  File "/media/pipx-3.8/venvs/poethepoet/lib/python3.8/site-packages/poethepoet/task/base.py", line 479, in run
    return self._handle_run(context, task_env)
  File "/media/pipx-3.8/venvs/poethepoet/lib/python3.8/site-packages/poethepoet/task/cmd.py", line 56, in _handle_run
    cmd = (*self._resolve_commandline(context, env), *extra_args)
  File "/media/pipx-3.8/venvs/poethepoet/lib/python3.8/site-packages/poethepoet/task/cmd.py", line 87, in _resolve_commandline
    for cmd_token, has_glob in resolve_command_tokens(
  File "/media/pipx-3.8/venvs/poethepoet/lib/python3.8/site-packages/poethepoet/helpers/command/__init__.py", line 59, in resolve_command_tokens
    for segment in word:
TypeError: 'Comment' object is not iterable
```

Is there something missing in my configuration in the above examples?

* Python version: **3.8**
* poethepoet version: **0.26.1**, installed with pipx
* pipx version: **1.5.0**

snejus:
Fixes #224


poethepoet/helpers/command/__init__.py
```diff
@@ -53,7 +55,8 @@ def finalize_token(token_parts):
         token_parts.clear()
         return (token, includes_glob)
 
-    for word in line:
+    non_comment_words = (w for w in line if not isinstance(w, Comment))
+    for word in non_comment_words:
```
nat-n, line 59:
```suggestion
    for word in line:
        if isinstance(word, Comment):
            # strip out comments
            continue
```

snejus: Your version is indeed more elegant

snejus:
Now regarding the test case with multiple comments - it actually evaluates to an invalid command if it's read literally (a comment between two flags does not make any sense).

Do you see any value in supporting this kind of thing? We could potentially pre-parse the command and ignore all comments altogether.

nat-n:
> Now regarding the test case with multiple comments - it actually evaluates to an invalid command if it's read literally (a comment between two flags does not make any sense).
> 
> Do you see any value in supporting this kind of thing? We could potentially pre-parse the command and ignore all comments altogether.

You're assessment is correct. Allowing a cmd to be split over multiple lines (with escaping the line break) with comments on multiple lines is a little bit weird... but harmless and IMO potentially a nice thing to be able to do. 

I think the root cause of the issue here is that although I made the AST implementation faithful to a subset of bash, when actually used in the cmd task it's configured to only treat `;` as a line terminator (not line breaks) and to later ignore lines with only whitespace or comments. However comments at the end of lines still end up terminating that line.

So as a solution I'm just pushed a tweak to the cmd AST that allows tracking on the `Line` object how it was terminated, so that we can implicitly merge lines that were only terminated with a comment. WDYT?

nat-n (approved):

nat-n:
This looks good to me now. Thanks for working on it @snejus

snejus:
Amazing, thanks for finishing this up!

Issue #223, opened 2024-06-15
User-defined variable substitution in [tools.poe.env] does not work since 0.26.0


dbohdan:
In Poe 0.25.1 I was able to substitute one `[tools.poe.env]` variable in another. Since 0.26.0, this no longer works. (Due to #214?)

Let's try the following `pyproject.toml` file.

```toml
[build-system]
requires = ["poetry-core"]
build-backend = "poetry.core.masonry.api"

[tool.poetry]
name = "test"
authors = []
description = ""
version = "0.0.1"

[tool.poe.env]
BASE_DIR = "/foo"
SUBDIR = "${BASE_DIR}/bar/"

[tool.poe.tasks.frobnicate]
cmd = "echo ${SUBDIR}"
```

Here is how the task `frobnicate` runs in different versions of Poe.

```none
> pipx install --force poethepoet==0.25.1
Installing to existing venv 'poethepoet'
  installed package poethepoet 0.25.1, installed using Python 3.12.1
  These apps are now globally available
    - poe
done! ✨ 🌟 ✨
> poe frobnicate
Poe => echo /foo/bar/
/foo/bar/
> pipx install --force poethepoet==0.26.0
[...]
done! ✨ 🌟 ✨
> poe frobnicate
Poe => echo /bar/
/bar/
> pipx install --force poethepoet==0.26.1
[...]
done! ✨ 🌟 ✨
> poe frobnicate
Poe => echo /bar/
/bar/
```

nat-n:
Hi @dbohdan, thanks for reporting this. 

Yep looks like something went wrong with 0.26.0. I'll look into this as soon as I have some time for it. 

If you able and willing to contribute a PR with a failing test case that would be a help :)

PR #227, opened 2024-06-18
Add failing test for #223

dbohdan:
I can confirm this test fails on Linux and Windows. :-)


nat-n:
@dbohdan Thanks for the help on this issue, turns out it was easy to fix :)

dbohdan:
Yay! You're welcome.

Issue #257, opened 2024-11-23
Support for `uv` package manager?


aaronsteers:
Our team uses Poe and Poetry together but we're planning to move to `uv` soon. Does Poe support `uv` and/or are there plans to support it in the future?

nat-n:
Hi @aaronsteers,
uv does look pretty neat, though I haven't yet worked with it myself. I suspect it wouldn't be too hard to make poethepoet work well with uv as well. 

In the best case it would be just a matter of implementing another executor class like [this one for poetry](https://github.com/nat-n/poethepoet/blob/main/poethepoet/executor/poetry.py) including some logic to locate the relevant venv for the current uv project and then tweaking [the base class](https://github.com/nat-n/poethepoet/blob/main/poethepoet/executor/base.py#L114) to be able to determine whether the current project uses uv.

I'd be interested to add support for this, though contributions – at least to validate and prototype the idea – would help.

nat-n:
I had a quick look at how uv manages virtual environments, and it turns out that since it does the conventional thing of keeping the the project venv directory at `./.venv` poe doesn't need to know anything about uv specifically to work with uv managed projected! I'll update the docs to clarify this.

So since it looks like **poethepoet already works with uv projects**, I'm going to close this issue. But I'd be interested to know if there are features missing to better integrate with a typical uv based workflow. For instance some accommodation may be required if/when uv adds support for [central management of venvs](https://github.com/astral-sh/uv/issues/1495).

aaronsteers:
Fantastic. Thanks, @nat-n!

AKuederle:
Maybe a small extension to this. `uv` recommends to always use `uv run` to run tasks within the project instead of activating the venv manually. The reason for this is, that `uv run` actually checks all packages are up to date. So first class support for `uv` could include doing that as well.

AKuederle:
I implemented a first version of a uv executor for poe in #271.

PR #271, opened 2025-01-12
Implemented first version of uv executor

AKuederle:
## Description of changes

This PR implements a uv-executor that uses `uv run` to execute tasks against the virtual env. The advantage over executing uv projects with the normal venv executor is, that `uv run` will automatically update and sync packages and is hence, recommended over activating the venv manually when working with uv.

@nat-n This is a first version, and I just wanted to get your feedback, if this is something you would be interested in merging.

See also #257 

## Pre-merge Checklist


tests/fixtures/uv_project/pyproject.toml
```diff
@@ -0,0 +1,21 @@
+[project]
+name = "uv-project"
+version = "0.1.0"
+description = "Add your description here"
+readme = "README.md"
+requires-python = ">=3.12"
+dependencies = []
```
nat-n, line 7: Do you think it's worth declaring a dependency here to exercise the integration more fully?

If yes there's a locally vendored package you can use at https://github.com/nat-n/poethepoet/blob/9997e878d4398b406c18ea7b23a49a176ec4c1a5/tests/fixtures/poetry_plugin_project/pyproject.toml#L11

nat-n:
Hey @AKuederle, this looks pretty good. Thanks for working on it!

AKuederle:
Yes I think the entire idea behind uv run is that running the additional checks is so fast, that it does not matter.
We might need to add a fallback (as I don't know what happens in cases where you don't have internet).

Regarding tests: I wasn't sure what tests to add. Based on my understanding, by adding a new project folder it should be included in the fixture. But I am not sure if this is sufficient. I also got some failing tests that did not seem to be related to my changes.

Could you provide some pointers, what I should update add, for you to consider merging this?

nat-n:
I'm curious how you would make an offline fallback work (if it's needed)?

I'm afraid the testing setup isn't quite that clever. Given the fixture project you created you'll need a test using [`run_poe_subproc`](https://github.com/nat-n/poethepoet/blob/main/tests/conftest.py#L115), for example: `run_poe_subproc("test-package-exec-version", project="uv")`.

Most tests work the same way, you can look here for example: https://github.com/nat-n/poethepoet/blob/main/tests/test_executors.py  You can either extend that test module or create a new more specific one.

If a test takes more than a couple of seconds to run (for example due to needed to setup the venv for the uv project) then it should be marked with `@pytest.mark.slow`, and if it requires the `uv` executable to be present to pass then it should be marked with `@pytest.mark.skipif(not shutil.which("uv"), reason="No uv available")`

I see a test failing due to a recent unreleased change I made. I pushed a fix to the development branch so you can rebase to get it.

nat-n:
I added a commit to complete the testing and documentation. Thanks again to @AKuederle for contributing this feature 😄

AKuederle:
Thank you for getting it over the finish line :) The initial prototype was not hard to do!

PR #311, opened 2025-06-21
Fix issue with locating sh.exe when git is installed in a custom loca…

per11235813:
Fix issue with locating sh.exe when git is installed in a custom location on windows

## Description of changes

An attempt to locate sh.exe on windows is relative to git.exe instead of just looking in `C:\Program Files\git`

This makes locating sh.exe work if git is installed in non-standard locations

## Pre-merge Checklist

- [ ] `poe check` executed successfully
- [ ] This PR targets the *development* branch for code changes or *main* if only documentation is updated

### DIFF
--- a/poethepoet/task/shell.py
+++ b/poethepoet/task/shell.py
@@ -1,6 +1,7 @@
 import re
 from collections.abc import Sequence
 from os import environ
+from pathlib import Path
 from typing import TYPE_CHECKING, Optional, Union
 
 from ..exceptions import ConfigValidationError, PoeException
--- a/poethepoet/task/shell.py
+++ b/poethepoet/task/shell.py
@@ -142,7 +142,7 @@ class ShellTask(PoeTask):
                 result = which(f"{prog_files}\\Git\\bin\\sh.exe")
 
                 if result is None and (git_path_str := which("git")) is not None:
-                    # Check if sh.exe can be found relative to the git executable 
+                    # Check if sh.exe can be found relative to the git executable
                     # in case git is installed at a non-standard location.
                     git_path = Path(git_path_str)
                     if (sh_path := git_path.parent.parent / "bin" / "sh.exe").exists():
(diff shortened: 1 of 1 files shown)


poethepoet/task/shell.py
```diff
@@ -141,6 +141,16 @@ def _locate_interpreter(self, interpreter: str) -> Optional[str]:
...
+            if result is None and (git_path_str := which("git")) is not None:
+                # git installation may be non-defualt library path to allow for 
+                # non-admin git-cli installation. In this case the posix shell
+                # can be located relative to the git executable.
+                from pathlib import Path
+
+                git_path = Path(git_path_str)
+                sh_path = git_path.parent.parent / "bin" / "sh.exe"
+                result = str(sh_path) if sh_path.exists() else None
+
```
nat-n, line 153: I don't have a windows box handy to test this so I'll have to trust you.

The the conditional logic doesn't look quite right to me. What do you about this?

```suggestion
            if result is None and self._is_windows:
                result = which(f"{prog_files}\\Git\\bin\\sh.exe")

                if result is None and (git_path_str := which("git")) is not None:
                    # Check if sh.exe can be found relative to the git executable 
                    # in case git is installed at a non-standard location.
                    git_path = Path(git_path_str)
                    if (sh_path := git_path.parent.parent / "bin" / "sh.exe").exists():
                        result = str(sh_path)
```

Also you might notice that I use lazy imports for performance reasons in many parts of this project but `pathlib` is an exception since it's used almost everywhere anyway.

per11235813: looks good. The attempt to find sh.exe relative to git.exe is now isolated to windows, which is probably a good thing.
I tested it on a clean windows 11 and it works.

nat-n (approved):

poethepoet/task/shell.py
```diff
@@ -141,6 +141,12 @@ def _locate_interpreter(self, interpreter: str) -> Optional[str]:
             if result is None and self._is_windows:
                 result = which(f"{prog_files}\\Git\\bin\\sh.exe")
 
+                if result is None and (git_path_str := which("git")) is not None:
+                    # Check if sh.exe can be found relative to the git executable
```
nat-n, line 145:
```suggestion
                    # Check if sh.exe can be found relative to the git executable
```

Issue #24, opened 2021-04-27
Is it possible to have a task  list execute in parallel


MartinWallgren:
Is it possible to create a task that executes multiple tasks in parallel?

I know I can create a compond task as

`test = ["mypy", "pylint", "pytest"]`

Calling `poe test` will run each task in sequence one after the other. It would be nice to be able to configure these task lists as safe to start in parallel.

Parallel should of course not be parallel by default since some tasks requires output from previous tasks (coverage being the prime example that needs a completed test run to before generating a coverage report).

nat-n:
This is not currently supported. I considered it when first implementing this sequence task type. I thought it might be nice if by default an array inside an array would be interpreted as a ParallelTask type within a SequenceTask, so for example the following would run mypy and pylint in parallel then pytest after that:

```toml
test = [["mypy", "pylint"], "pytest"]
```

And of course you could also do:

```toml
test.parallel = ["mypy", "pylint", "pytest"]
```

However the problem is that I'm sure what it should do with stdout. I imagine one wouldn't simply want both subprocesses to write to the same console at the same time! Maybe there could be a solution along the lines of capturing the output and feeding it out to the console one line at a time (maybe with a prefix linking it to the task that produced it) but that's getting complicated to implement.

As I mention in #26, if the stdout of those tasks were configured to be captured anyway – such as for use in another task, or maybe to be piped to a file or discarded – then this problem goes away, and the tasks might as well be run in parallel. There's just the question left of how to handle a failure of one task in the set (whether to wait for the others).

I'd like to support parallel execution, but I'm really not sure how it should work. What do you think?

nat-n:
Also a potential if inelegant workaround might be to use a shell task with background jobs, like something along the lines of:

```toml
[tool.poe.tasks.test]
shell = """
poe mypy &
poe pylint &
poe pytest &
wait $(jobs -p)
"""
```

asfaltboy:
We do something like this in bash:

1. First run all processes together, use a temp file to store each command's output
2. Then, we iterate over each command waiting for its completion and display that command's output.

The side effect of this simple method is that it seemingly "stalls" on the slowest command, returning when they all complete. This means that CMDS array should preferably be sorted fastest to slowest.

<details>
<summary>Click to expand!</summary>

```bash
for cmd in "${CMDS[@]}"; do
    stdout="$(mktemp)"
    timer="$(mktemp)"
[...]

PR #316, opened 2025-07-14
Allow tasks to run in parallel

sewi-cpan:
The `sequence` executes tasks one-by-one using a single CPU core unless the called command can use more.

The new `parallel` option runs the specified tasks concurrently. Each `parallel` task group waits untill all subtasks finished.

## Description of changes

See issue #24 for the discussion of this feature.

### STDOUT and STDERR

> [However the problem is that I'm not sure what it should do with stdout. I imagine one wouldn't simply want both subprocesses to write to the same console at the same time!](https://github.com/nat-n/poethepoet/issues/24#issuecomment-827917083)

I spent a few days trying to capture the STDOUT and STDERR of subtasks failing on multiple problems.

**1. Set `capture_output` - but subtasks can be anything**

Setting the `capture_stdout` argument to a temporary file is easy - if the next task supports it! Subtasks are usually `ref` tasks, eventually ending in a command. The `parallel` task implementation would need to walk down the tree and modify any subtasks to capture the output reliably.

This is too hacky.

**2. Use threads - but STDOUT and STDERR are common to the process**

Each subtask can easily run in a dedicated thread (or using a pool). The Poe source code itself can be changed to accept an alternate file handle for output of a task, while the output of external processes still goes to STDOUT and STDERR.

Changing the file descriptors to redirect STDOUT and STDERR doesn't work because every thread would change the same shared file descriptors.

Example (STDERR behaves the same as STDOUT):
* Thread 1 changes STDOUT to `/tmp/task1.txt`
* Thread 2 changes STDOUT to `/tmp/task2.txt`
* Thread 1 executes a shell command
* The command of thread 1 outputs something

The output ends up in `/tmp/task2.txt`.

**3. `fork` to have separate file descriptors - doesn't work on Windows**

Linux supports a `fork` kernel call, cloning a running process into two. This is a perfect solution: The new clone continues to run at the same source code location as the parent. The clone's STDOUT and STDERR are independent from the parent process and can easily be redirected. A `CHLD` signal notifies the parent process when the clone finishes, and Poe has no database connections that might be corrupted by forking.

Creating processes is different on Windows. Nothing like `fork` exists. Poe must run on all operating systems.

**4. [The `ProcessPoolExecutor` works on Windows](https://docs.python.org/3/library/concurrent.futures.html#processpoolexecutor) - but not in Poe**

Poe is using IO streams within the context. The `ProcessPoolExecutor` must be able to serialize the arguments and fails on the Poe `RunContext`.

**5. Start another Poe process using the shell - but Windows and arguments**

`poe subtask >/tmp/out 2>/tmp/err` is hacky. Working around the differences between Windows (no STDERR) and other OS is possible.

The command line arguments and the configuration context can't be passed easily.

**Conclusion**

Most of the attempts can be done for the price of a huge refactoring of unrelated Poe parts. I decided not to do this in the first implementation.

The `capture_stdout` option is available to split the output of each subtask. The output of all tasks is now mixed by default. Unix-style OS users can easily append `| sed 's/^Prefix: /'` to each command if they want.

**Behavior on failing tasks**

> [There's just the question left of how to handle a failure of one task in the set (whether to wait for the others).](https://github.com/nat-n/poethepoet/issues/24#issuecomment-827917083)

Running tasks in parallel instead of sequentially is a huge performance boost. Killing concurrently running tasks if one fails is risky if a process is killed during a write operation (e.g. a reformatter writing the formatted changes of a file). Read-only processes can be killed safely.

The current implementation treats subtasks as atomic:
* Wait for running subtasks to complete
* Silently discard subtasks that haven't been started

## Pre-merge Checklist

- [X] New features (if any) are covered by new feature tests
- [X] New features (if any) are documented
- [X] Bug fixes (if any) are accompanied by a test (if not too complicated to do so) - no bug fixes included
- [X] `poe check` executed successfully
- [X] This PR targets the *development* branch for code changes or *main* if only documentation is updated

Transparency note: Co-authored by Claude. All AI output has been reviewed before creating the PR.


sewi-cpan:
The CI test failures seem to be unrelated. What to do now?

sewi-cpan:
> 2. I initially had in mind to execute parallel tasks using `asyncio.create_subprocess`, which I have a feeling might provide an easier way to get around the issue of managing output streams. What do you think? Unfortunately it probably requires a larger refactor of how PoeExecutors work (and probably also how tasks call one another, I was thinking about a centralized async executor dispatch via event passing), which is why I've been putting it off for a while... If the threading based approach works well enough for now then that's OK.

Using `asyncio` causes the problem 2 mentioned above: Each process has only one STDOUT/STDERR pair. Redirecting that for a subtask overwrites all previous redirections for other subtasks.

Likely, `asyncio` is the best solution if
* each task knows the IO streams to be used for STDOUT/STDERR
* All internal output (`print`) uses them directly
* the section {redirect STDOUT/STDERR, launch external command, revert redirect} is atomic (critical section that isn't interrupted

Each child process has it's own file descriptors for STDOUT and STDERR. Changing them (back) in the Poe process is safe after the child has been started. A `parallel_cmd` task type (having commands as config options instead of other tasks) would be easy:
```Python
children = [ Popen(cmd, stdout=PIPE, stderr=PIPE) for cmd in cmd_list ]
for child in children:
  child.wait()
  print(child.stdout)
  print(child.stderr)
```
(Mockup code, not real.)
...but that would break the Poe concept of nested, re-usable tasks.

I'm on a project using Poe for pre-commit clean up. A lot of those tasks can run in parallel and have no or very little output on success and huge error messages on failure (for example, `mypy`, `docformatter`, `makemigrations` - each of them running multiple times on different sub-parts). Each pre-commit check takes ~5 minutes with 100% load on one CPU core. Running them in parallel would help me a lot. Managing the output is nice-to-have for me. Other people will have different priorities.

nat-n:
I had a deeper look and I think a few things need fixing before this can be merged (see inline comments).

Also some more rigorous testing is required. I'd suggest the following test strategy:
1. Create a dedicated test project for parallel task tests (or [generate one dynamically](https://github.com/nat-n/poethepoet/blob/main/tests/test_ignore_fail.py#L5)).
2. Create a test scenario that actually requires parallel task execution for it to pass. This could be done by creating a new helper task that delays some ms to use instead of poe_test_echo like `"poe_test_delayed_echo 50 'Running task'"`
3. Test for some variety of tasks, e.g. 0 or 1, 25 subtasks, incline cmd task, etc
4. Test a parallel task with a failing subtask, with or without ignore_fail configured
5. test default_item_type

If you're keen to carry this PR through to completion then great, otherwise I'll come finish it off when I finish with other changes I have in progress.

PR #326, opened 2025-11-06
feat: allow optional envfiles without warnings

### LINKED ISSUE #321: Allow tool.poe.envfile to be missing without a warning
In the 0.36.0 version, I could define:
```
[tool.poe]
envfile = ".env"
shell_interpreter = "bash"
```
Into my `pyproject.toml` file and without needing to use `POE_DEBUG` or:
```
[tool.poe]
envfile = ".env"
shell_interpreter = "bash"
verbosity = -1
```
I could see each task that ran a shell command print the command in blue before it ran, but I didn't see the `Warning: Poe failed to locate envfile at ...`. Now in v0.37.0 I am forced to either lose the blue text, view the warning, or manually ensure there is a `.env` file present to get the same output.
I'd love to see a setting where `tool.poe.envfile` can be optional without raising a warning. I would also like to retain the ability to print the command before it is run while using this setting.

cnaples79:
## Summary
- accept optional/expect syntax for global and task  definitions
- skip the warning when optional envfiles are missing while still loading required files in order
- document the table form and add regression tests covering both cases

## Testing
- ============================= test session starts ==============================
platform linux -- Python 3.12.12, pytest-8.4.2, pluggy-1.6.0
rootdir: /root/projects/GitHub-OSS/poethepoet
configfile: pyproject.toml
plugins: anyio-4.11.0
collected 2 items

tests/test_envfile_optional.py ..                                        [100%]

============================== 2 passed in 2.11s ===============================

Fixes #321

### DIFF
--- a/docs/global_options.rst
+++ b/docs/global_options.rst
@@ -50,17 +50,17 @@ You can configure environment variables to be set for all poe tasks in the pypro
   VAR1 = "FOO"
   VAR2 = "BAR BAR BLACK ${FARM_ANIMAL}"
 
-The example above also demonstrates how – as with env vars defined at the task level –
-posix variable interpolation syntax may be used to define global env vars with reference
-to variables already defined in the host environment or in a referenced env file.
+The example above also demonstrates how – as with env vars defined at the task level – posix variable interpolation syntax may be used to define global env vars with reference to variables already defined in the host environment or in a referenced env file.
 
-As with the task level option, you can indicated that a variable should only be set if
-not already set like so:
+As with the task level option, you can indicated that a variable should only be set if not already set like so:
 
 .. code-block:: toml
 
   [tool.poe.env]
   VAR1.default = "FOO"
+
+Loading external environment variables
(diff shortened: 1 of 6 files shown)


tests/test_envfile_optional.py
```diff
@@ -0,0 +1,61 @@
...
+
+def test_optional_envfile_missing(run_poe_subproc, projects, tmp_path):
+    project_path = _copy_fixture_project(projects, "envfile_optional", tmp_path)
+
+    optional_env = project_path / ".env"
+    if optional_env.exists():
+        optional_env.unlink()
+
+    result = run_poe_subproc(
+        "print-var",
```
nat-n: if you do this (or set `NO_COLOR` in the env) then there's no need to stub Pastel.
```suggestion
        "--no-ansi", ""print-var",
```

xenador1:
Looks like https://github.com/nat-n/poethepoet/pull/329 solves the need for Python 3.9

I'm really looking forward to trying this out when it releases. Thank you for putting in the effort to make it a reality.

PR #376, opened 2026-04-12
fix: preserve quotes in :+/:- operation arguments

### LINKED ISSUE #333: Alternate-value operator strips quotes from substituted value
When using the alternate-value operator (`${VAR:+...}`), the value provided in the substitution is not passed through correctly. Specifically, quote characters are being dropped somewhere in the evaluation pipeline, causing the final command to receive an unquoted version of the intended string.

This leads to unexpected argument splitting and incorrect command behaviour.

```
[tool.poe.tasks.reproduction]
args = [
    { name = "SKIP_BUILD", options = ["--skip-build"], type = "boolean" },
]
cmd = "echo pytest ${SKIP_BUILD:+ -m 'not build'}"
```

The quotes around "not build" are removed, so it is interpreted as two separate arguments rather than one.
```
Poe => echo pytest -m not build
pytest -m not build 
```

Losing the quotes breaks scenarios where the substituted value must remain a single token, such as pytest markers.

lawrence3699:
Fixes #333

## Problem

When using parameter expansion operators like `${VAR:+ -m 'not build'}`, the
operation argument is resolved to a flat string via `resolve_param_argument`
before being word-split. This strips the quote structure, causing single-quoted
(or double-quoted) multi-word values to be split into separate command tokens.

**Before:** `echo pytest ${SKIP_BUILD:+ -m 'not build'}` with `SKIP_BUILD` set
produces tokens `echo`, `pytest`, `-m`, `not`, `build` (5 tokens — quotes lost).

**After:** Same expression produces tokens `echo`, `pytest`, `-m`, `not build`
(4 tokens — quoted string preserved as a single token).

## Fix

When a `ParamExpansion` in an unquoted context has an active `:+` or `:-`
operation, expand the operation's `ParamArgument` segments inline within
`resolve_command_tokens`, respecting each segment's quote boundaries:

- Quoted segments → content stays as a single token part (no word splitting)
- Unquoted segments → `WhitespaceText` nodes produce token boundaries as usual

This replaces the previous path that flattened the entire argument to a string
before word splitting.

## Validation

- All 16 existing `tests/helpers/command/` tests pass
- All 18 existing `tests/test_cmd_param_expansion.py` tests pass
- Added regression test covering single quotes in `:+`, double quotes in `:-`,
  and mixed quoted/unquoted content in operation arguments

### DIFF
--- a/poethepoet/helpers/command/__init__.py
+++ b/poethepoet/helpers/command/__init__.py
@@ -34,7 +34,14 @@ def resolve_command_tokens(
     patterns that are not escaped or quoted. In case there are glob patterns in the
     token, any escaped glob characters will have been escaped with [].
     """
-    from .ast import Glob, ParamArgument, ParamExpansion, ParseConfig, PythonGlob
+    from .ast import (
+        Glob,
+        ParamArgument,
+        ParamExpansion,
+        ParseConfig,
+        PythonGlob,
+        WhitespaceText,
+    )
 
     if not config:
         config = ParseConfig(substitute_nodes={Glob: PythonGlob})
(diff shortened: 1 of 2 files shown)


nat-n:
HI @lawrence3699, thanks for picking this issue up!

As it happens I had also started looking into it a few days ago and got as far as writing some failing tests (the low effort part to automate :D) https://github.com/nat-n/poethepoet/pull/377 I suggest rebasing to use those tests (which subsume and extend the one's you provided).

The code needs some iteration. The Copilot PR comments are valid, but also diff adds some complexity with a new parallel processing code path instead of fixing the root cause of the issue (resolve_param_argument should return tokens instead of a pre-joined string). Also the resolve_command_tokens is getting a bit long with excessive indentation and repetition, some code cleanup is required (it was already at the limit).

nat-n:
Closing this PR in favour of the more complete solution in #377

The fix will be available in the next release.

PR #375, opened 2026-04-09
Passing free arguments to sequence, parallel and switch tasks is now possible.

timrid:
## Description of changes
This PR adds the possibility to pass free arguments to sequence and parallel tasks. These free arguments are passed to every subtask.

The motivation for this is described in #374.

Fixes #374

## Pre-merge Checklist

- [x] New features (if any) are covered by new feature tests
- [x] New features (if any) are documented
- [ ] Bug fixes (if any) are accompanied by a test (if not too complicated to do so)
- [x] `poe check` executed successfully
- [x] This PR targets the *development* branch for code changes or *main* if only documentation is updated

PR #378, opened 2026-04-13
fix: handle cancelled tasks correctly

timrid:
## Description of changes
When working on https://github.com/nat-n/poethepoet/pull/375 I had a sporadic CI error in `test_parallel_fail_all` that was unrelated to my changes (see [this pipeline run](https://github.com/nat-n/poethepoet/actions/runs/24188199358/job/70935014348?pr=375)). So I started investigating and found this error:

According to [the `asyncio.Task.exception` docs](https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.exception)

> If the Task has been cancelled, this method raises a CancelledError exception.

So it is necessary to check for cancelled task before accessing the exception object.

With this change I cannot reproduce this error on my Mac in 100 pytest runs with

```
for i in $(seq 100); do pytest tests/test_parallel_tasks.py::test_parallel_fail_all -x || break; done
```

So I removed the flaky marker for this test. Maybe some other flaky markers are now also obsolet, but `test_parallel_fail_all` was the only one I could reproduce locally before my change and cant reproduce after it.

## Pre-merge Checklist

### DIFF
--- a/poethepoet/executor/task_run.py
+++ b/poethepoet/executor/task_run.py
@@ -200,17 +200,18 @@ class PoeTaskRun:
     async def _watch_completion(self) -> None:
         await self.wait()
 
-        if not self.has_failure:
-            event = PoeTaskRunCompletion(self.name)
-        else:
-            event = PoeTaskRunError(
+        event = (
+            PoeTaskRunCompletion(self.name)
+            if not self.has_failure
+            else PoeTaskRunError(
                 self.name,
                 exception=(
                     None
                     if self.asyncio_task.cancelled()
                     else self.asyncio_task.exception()
-                )
+                ),
             )
+        )
 
         await self._notify_and_clear_done_callbacks(event)
         self._clear_completion_watcher()
--- a/poethepoet/executor/task_run.py
+++ b/poethepoet/executor/task_run.py
@@ -200,11 +200,17 @@ class PoeTaskRun:
     async def _watch_completion(self) -> None:
         await self.wait()
 
-        event = (
-            PoeTaskRunCompletion(self.name)
-            if not self.has_failure
-            else PoeTaskRunError(self.name, exception=self.asyncio_task.exception())
-        )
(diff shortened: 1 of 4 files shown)


tests/test_parallel_tasks.py
```diff
@@ -163,7 +163,6 @@ def fmt_ignore_fail(value):
     return generator
 
 
-@pytest.mark.flaky(reruns=3, reruns_delay=1)
```
nat-n, line 166: Looks like the test is still flaky on lesser runtimes :( https://github.com/nat-n/poethepoet/actions/runs/24363356470/job/71164123837?pr=378

poethepoet/executor/task_run.py
nat-n: Looks like we need the same fix here actually:

```suggestion
            else PoeTaskRunError(
                self.name, 
                exception=(None if self.asyncio_task.cancelled() else self.asyncio_task.exception())
            )
```

Maybe the whole statement doesn't fit nicely in a ternary expression anymore but you get what I mean.

nat-n (changes requested):
Thanks for the fix :)

poethepoet/executor/task_run.py
timrid: Done

timrid:
I ran the CI a few more times in my fork and found some more issues I fixed. See for example [here](https://github.com/timrid/poethepoet/actions/runs/24420801559/job/71342103350) or [here](https://github.com/timrid/poethepoet/actions/runs/24422385728/job/71347794809).

To manually trigger the CI I also added `workflow_dispatch`. Maybe this is also useful for you, so I added this change to this PR.

Now I ran the CI 3 times without errors.

tests/test_parallel_tasks.py
timrid: Fixed some more flakyness. Now the CI passes 3 times without errors.

PR #380, opened 2026-04-17
feat: add support for forwarding free arguments via $POE_EXTRA_ARGS

### LINKED ISSUE #155: No way to pass free arguments to a specific place
With shell scripts, it's possible to do something like: `python -m robot --pythonpath . -d ./outptut "$@" tests`, so that when you call a script, the passed arguments are forwarded to the command in a specific location.

However with poe a task like: ```toml
[tool.poe.tasks.test]
help = "Run all tests"
cmd = "python -m robot --pythonpath . -d ./output ./tests"
```

Any passed arguments will simply be appended to the end of the command. This isn't ideal, as some CLI tools only support usage like: `robot [options] paths`, which means any positional arguments left after the options are interpreted as paths. This means that running for example: `robot --pythonpath . ./tests -d ./output` wouldn't work, and I'd get an error that it couldn't find `-d` as a path.

I'd love to see a feature that would allow for me to pass arguments to a specific place in my command.

### LINKED ISSUE #374: Free arguments in sequence and parallel tasks
**Problem description:**
I would like to set up a tox-like configuration with poethepoet as shown in the example from the documentation:

```
[tool.poe.executor]
type = "uv" # Use the uv executor for all tasks by default

[tool.poe.tasks.test-py39]
cmd = "pytest tests --cov=mypackage"
executor = {isolated = true, python = "3.9"} # Specify python version, and that we want an isolated env

[tool.poe.tasks.test-py310]
cmd = "pytest tests --cov=mypackage"
executor = {isolated = true, python = "3.10"}

[tool.poe.tasks.test-py311]
cmd = "pytest tests --cov=mypackage"
executor = {isolated = true, python = "3.11"}

[tool.poe.tasks.test-py312]
cmd = "pytest tests --cov=mypackage"
executor = {isolated = true, python = "3.12"}

[tool.poe.tasks.test-all]
sequence = ["test-py39", "test-py310", "test-py311", "test-py312"]
```

With this configuration I can then run all tests with:

`poe test-all`

However, I would prefer a more flexible setup where I can pass `--cov=mypackage` and more args as a free argument, like this:

```
[tool.poe.executor]
type = "uv" # Use the uv executor for all tasks by default

[tool.poe.tasks.test-py39]
cmd = "pytest tests"
executor = {isolated = true, python = "3.9"} # Specify python version, and that we want an isolated env

[tool.poe.tasks.test-py310]
cmd = "pytest tests"
executor = {isolated = true, python = "3.10"}

[tool.poe.tasks.test-py311]
cmd = "pytest tests"
executor = {isolated = true, python = "3.11"}

[tool.poe.tasks.test-py312]
cmd = "pytest tests"
executor = {isolated = true, python = "3.12"}

[tool.poe.tasks.test-all]
sequence = ["test-py39", "test-py310", "test-py311", "test-py312"]
```

and then call it like:

`poe test-all --cov=mypackage`

This makes it way more flexible to temporarily include more flags for all tests e.g. when using verbose mode by `-v`:

`poe test-all --cov=mypackage -v`

But currently, free arguments are not forwarded to the individual sequence tasks. 

**Requested feature:**
Passing free arguments to a `sequence` task should forward those arguments to the individual tasks in the sequence.

Additionally, the same problem described above applies to the `parallel` task as well, where free arguments should also be forwarded to the individual tasks being run in parallel.

### PR DESCRIPTION
**Pull request by timrid**
## Description of changes

When a task receives free arguments — either positional arguments not matched by any named arg definition, or arguments passed after `--` — poe now sets a `$POE_EXTRA_ARGS` environment variable containing those arguments as a shell-quoted, space-delimited string. This variable is inherited by all subtasks and can be referenced anywhere environment variable expansion is supported.

Previously, cmd tasks silently appended extra args to the end of the command and most other task types discarded them. `$POE_EXTRA_ARGS` gives full control over placement — it can appear anywhere in a subtask string, including between fixed options:

```
[tool.poe.tasks.test-all]
sequence = [
  "pytest --timeout=30 $POE_EXTRA_ARGS --tb=short",
  "lint",
]
```

For `script` and `expr` tasks, free args are also available as `_extra_args` (a `list[str]`) directly inside the Python expression:

```
[tool.poe.tasks.run-tests]
script = "pytest:main(_extra_args)"
```

For `cmd` tasks, referencing `$POE_EXTRA_ARGS` explicitly opts out of the default auto-append behavior so extra args are only placed where specified.

(Note: This code was largely generated with AI assistance and manually reviewed.)

Fixes #155 
Fixes #374 

## Pre-merge Checklist

### DIFF
--- a/poethepoet/task/parallel.py
+++ b/poethepoet/task/parallel.py
@@ -172,13 +172,6 @@ class ParallelTask(PoeTask):
         named_arg_values, extra_args = self.get_parsed_arguments(env)
         env.register_task_args(named_arg_values, extra_args)
 
-        if (
-            not named_arg_values
-            and not extra_args
-            and any(arg.strip() for arg in self.invocation[1:])
-        ):
-            raise PoeException(f"Parallel task {self.name!r} does not accept arguments")
-
         if len(self._subtasks) > 1:
             # Indicate on the global context that there are multiple stages
             context.multistage = True
--- a/poethepoet/task/parallel.py
+++ b/poethepoet/task/parallel.py
@@ -169,10 +169,14 @@ class ParallelTask(PoeTask):
     async def _handle_run(
         self, context: RunContext, env: TaskEnv, task_state: PoeTaskRun
     ):
-        named_arg_values, _ = self.get_parsed_arguments(env)
-        env.register_task_args(named_arg_values)
+        named_arg_values, extra_args = self.get_parsed_arguments(env)
+        env.register_task_args(named_arg_values, extra_args)
 
-        if not named_arg_values and any(arg.strip() for arg in self.invocation[1:]):
+        if (
+            not named_arg_values
+            and not extra_args
+            and any(arg.strip() for arg in self.invocation[1:])
+        ):
             raise PoeException(f"Parallel task {self.name!r} does not accept arguments")
 
         if len(self._subtasks) > 1:
--- a/poethepoet/task/ref.py
+++ b/poethepoet/task/ref.py
@@ -88,7 +88,7 @@ class RefTask(PoeTask):
         invocation_tokens = tuple(
             env.fill_template(token) for token in shlex.split(expanded_content)
         )
-        if "POE_EXTRA_ARGS" in self.spec.content:
+        if self._content_uses_extra_args():
             ref_invocation = invocation_tokens
         else:
             ref_invocation = (*invocation_tokens, *extra_args)
--- a/poethepoet/task/ref.py
+++ b/poethepoet/task/ref.py
@@ -82,15 +82,16 @@ class RefTask(PoeTask):
             task_state.ignore_failure(ignore_fail)
 
         named_arg_values, extra_args = self.get_parsed_arguments(env)
-        env.register_task_args(named_arg_values)
+        env.register_task_args(named_arg_values, extra_args)
 
-        ref_invocation = (
-            *(
-                env.fill_template(token)
-                for token in shlex.split(env.fill_template(self.spec.content.strip()))
-            ),
-            *extra_args,
+        expanded_content = env.fill_template(self.spec.content.strip())
+        invocation_tokens = tuple(
+            env.fill_template(token) for token in shlex.split(expanded_content)
         )
+        if "POE_EXTRA_ARGS" in self.spec.content:
+            ref_invocation = invocation_tokens
+        else:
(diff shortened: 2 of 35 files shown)

### CODE REVIEW

File: poethepoet/task/parallel.py [hunk 1]
```diff
@@ -169,10 +169,14 @@ def __init__(
...
+        named_arg_values, extra_args = self.get_parsed_arguments(env)
+        env.register_task_args(named_arg_values, extra_args)
 
-        if not named_arg_values and any(arg.strip() for arg in self.invocation[1:]):
+        if (
+            not named_arg_values
+            and not extra_args
+            and any(arg.strip() for arg in self.invocation[1:])
+        ):
             raise PoeException(f"Parallel task {self.name!r} does not accept arguments")
```
nat-n: I don't think these guards make sense anymore, same for sequence, shell, and switch.
```suggestion
```

timrid: You are right, I removed these Exceptions.

poethepoet/task/ref.py
```diff
@@ -82,15 +82,16 @@ async def _handle_run(
...
-                for token in shlex.split(env.fill_template(self.spec.content.strip()))
-            ),
-            *extra_args,
+        env.register_task_args(named_arg_values, extra_args)
+
+        expanded_content = env.fill_template(self.spec.content.strip())
+        invocation_tokens = tuple(
+            env.fill_template(token) for token in shlex.split(expanded_content)
         )
+        if "POE_EXTRA_ARGS" in self.spec.content:
```
nat-n, line 91: This test could be better... at least `"$POE_EXTRA_ARGS" in self.spec.content or "${POE_EXTRA_ARGS" in self.spec.content` ?

The proper solution I'd like to see (not a blocker) depends on a more complex refactor I just started on to use the bash ast parser for all env var templating, so I'll come back to this later anyway.

timrid: > This test could be better... at least `"$POE_EXTRA_ARGS" in self.spec.content or "${POE_EXTRA_ARGS" in self.spec.content` ?

Done

nat-n (approved):

timrid:
@nat-n Thanks for merging this. Do you have a time schedule for this to get released?

nat-n:
> @nat-n Thanks for merging this. Do you have a time schedule for this to get released?

Shipped in [v0.45.0](https://github.com/nat-n/poethepoet/releases/tag/v0.45.0) 🚀

PR #389, opened 2026-05-18
refactor: Extend PoeOptions with validation constraints and descriptions

nat-n:
## Description of changes

Phase 1 of the [JSON Schema generation work](docs/superpowers/specs/2026-05-17-poe-jsonschema-generation-design.md): refactor `PoeOptions` to be more declarative so a future schema generator (Phase 2) can read most of what it needs directly from annotations. This is a foundation PR — no JSON Schema is emitted yet.

Three threads of work, kept as independent commits for review:

**1. Literal-tighten runtime-validated string options.**
- New `register_type_alias` helper in `poethepoet/options/annotations.py` lets non-class type aliases (Literals, Unions) be referenced by name in `PoeOptions` annotations.
- `ShellInterpreter` (a `Literal[...]`) replaces the parallel `KNOWN_SHELL_INTERPRETERS` tuple — the tuple is now derived via `get_args(ShellInterpreter)` so the two stay in lockstep.
- `ShellTask.TaskOptions.interpreter` and `ProjectConfig.ConfigOptions.shell_interpreter` are retyped to use the alias; their bespoke `validate()` value-membership checks are deleted (now handled declaratively by `LiteralType.validate`).

**2. Extend `Metadata` with declarative constraint fields.**
- Adds `pattern`, `examples`, `minimum`, `maximum`, `min_length`, `max_length` to `poethepoet.options.annotations.Metadata`.
- All constraints (except documentation-only `examples`) are enforced at parse time in `PrimitiveType.validate` / `ListType.validate`, with messages that name the field, value, and constraint.
- `pattern` uses `re.search` (unanchored) to match JSON Schema's `pattern:` semantics.
- `min_length=1` applied to `ShellTask.TaskOptions.interpreter` restores the empty-list rejection that the Literal refactor temporarily removed.
- Includes a fix to `TypeAnnotation.metadata_get` so falsy-but-set metadata values (`0`, `False`, `""`) are no longer silently dropped.

**3. AST-based class-attribute docstring extraction.**
- New `PoeOptions.description_for_field(name)` reads PEP 257-style class-attribute docstrings via `ast`, walks MRO so inherited fields resolve to the nearest ancestor that defines them, caches per class.
- Every annotated field on every `PoeOptions` / `ExecutorOptions` / `ArgSpec` / `IncludedConfig.ConfigOptions` / `PackagedConfig.ConfigOptions` / TypedDict has a docstring backfilled (sourced from the existing schemastore `partial-poe.json` where applicable, fresh where not).
- Comprehensiveness tests (`tests/options/test_descriptions.py`) walk each registry and fail if any field is missing a description — adding a new option without a docstring will fail CI.

**Why this matters:** Phase 2 will introduce `poethepoet/schema/` to auto-generate `partial-poe.json` from these annotations. By making the source of truth declarative (Literals + Metadata + docstrings) rather than scattered across bespoke `validate()` overrides, the schema generator becomes a translator instead of an interpreter, and the design principle "schema constraints must be real runtime rules" is structurally enforced.

Design doc: `docs/superpowers/specs/2026-05-17-poe-jsonschema-generation-design.md`
Implementation plan: `docs/superpowers/plans/2026-05-17-poe-jsonschema-phase1.md`

Two architectural changes worth flagging for review:
- `UnionType.__init__` now propagates the union's metadata to child types when the child has no metadata of its own — needed so `Annotated[X | Sequence[X] | None, Metadata(min_length=1)]` reaches `ListType.validate`. Verified safe for the existing `Metadata(config_name=...)` uses on `executor/uv.py` and `task/expr.py` (config_name is consumed at field level, not by per-branch validation).
- `register_type_alias`'s docstring recommends the split form (`Foo = Literal[...]; register_type_alias("Foo", Foo)`) over the inline form, because the inline form types `Foo` as `Any` for mypy, breaking same-module annotation use. The split form works in both contexts.

## Pre-merge Checklist

- [x] Bug fixes (if any) are accompanied by a test (if not too complicated to do so)
- [x] \`poe check\` executed successfully (1020 passed, 3 skipped)
- [x] This PR targets the *development* branch

### DIFF
--- a/poethepoet/config/partition.py
+++ b/poethepoet/config/partition.py
@@ -39,8 +39,21 @@ KNOWN_SHELL_INTERPRETERS: tuple[str, ...] = get_args(ShellInterpreter)
 @option_annotation
 class IncludeScriptItem(TypedDict):
     script: str
+    """
+    A reference to a Python callable that returns toml or json-like config to be
+    merged into the project config.
+    """
+
     cwd: str
+    """
+    Specify the working directory for resolving relative paths referenced by the
+    included script.
+    """
+
     executor: str | dict
+    """
+    Configure the executor used when invoking the include_script callable.
+    """
 
 
 IncludeScriptItem.__optional_keys__ = frozenset({"cwd", "executor"})
@@ -49,8 +62,21 @@ IncludeScriptItem.__optional_keys__ = frozenset({"cwd", "executor"})
 @option_annotation
 class IncludeItem(TypedDict):
     path: str
+    """
+    The path to the toml or json config file to include, relative to the parent
+    config file.
+    """
+
     cwd: str
+    """
+    Specify the working directory for resolving relative paths referenced by the
+    included config.
+    """
+
     recursive: bool
+    """
+    If true, includes declared in the included file will also be loaded.
+    """
 
 
 IncludeItem.__optional_keys__ = frozenset({"cwd", "recursive"})
@@ -59,8 +85,19 @@ IncludeItem.__optional_keys__ = frozenset({"cwd", "recursive"})
 @option_annotation
 class TaskGroup(TypedDict):
     heading: str
+    """
+    A human-readable name for the group displayed in the help output.
+    """
+
     executor: Mapping[str, str | Sequence[str] | bool] | str | None = None  # type: ignore[misc]
+    """
+    Configure the executor used by tasks within this group.
+    """
+
     tasks: Mapping[str, Any] = EmptyDict  # type: ignore[misc]
+    """
+    The tasks defined within this group.
+    """
 
 
 TaskGroup.__optional_keys__ = frozenset({"executor", "tasks"})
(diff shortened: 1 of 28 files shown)


nat-n (commented):
## Review — JSON Schema Phase 1

Strong foundational work. The `metadata_get` falsy-fix is correct and well-tested, the AST docstring extraction is defensively coded, and the comprehensiveness tests give real teeth to the "every field is documented" guarantee. One architectural concern worth resolving before Phase 2 lands, plus a few smaller items.

### 1. UnionType metadata propagation — surprising semantics ahead

The author's safety claim for the four current sites checks out: the three `Metadata(config_name=...)` propagations (`executor/uv.py:35,41`, `task/expr.py:35`) are benign-but-pointless because `config_name` is only read at field level, and the `min_length=1` on `task/shell.py:34` works because only the `Sequence[...]` branch in `ShellInterpreter | Sequence[ShellInterpreter] | None` honours `min_length`. So today, fine.

The risk is the **next** contributor adding a field shaped like `str | list[str]` with `Metadata(min_length=N)`. JSON Schema deliberately separates `minLength` (string characters) from `minItems` (list items) — the runtime currently conflates them because each constraint is only enforced on one type kind. Same trap for `pattern` propagating to a Literal child (silently ignored), and for numeric bounds reaching a `bool | int` field (since `bool` is an `int` subclass).

Recommendations, any of which would address this:
- Make propagation **field-selective** — whitelist `min_length`/`max_length`/`minimum`/`maximum`/`pattern`/`examples`, explicitly exclude `config_name`.
- Better: leave propagation off and require `Annotated[]` on the relevant branch.
- At minimum: add a test for `str | list[str]` + `Metadata(min_length=N)` to lock in current behaviour, and a comment in `UnionType.__init__` documenting the constraint.

### 2. `_wrap_with_metadata` mutates `child._metadata` post-construction

`annotations.py:380-382` mutates state on an already-constructed `TypeAnnotation`. Currently safe (every `TypeAnnotation.parse(arg)` returns a fresh instance), but it breaks an immutability invariant suggested by the surrounding design. If `parse` ever caches — or anyone makes `_metadata` a frozen dataclass — this becomes a foot-gun. Worth a comment, or constructing the metadata-wrapped child in one step.

### 3. `register_type_alias` collision detection

Silent overwrite of `_registered_type_hint_globals[name]`. Same pre-existing flaw as `option_annotation`, but the surface area is growing. A `raise` (or at least a warning) on duplicate registration would catch refactor mistakes early.

### 4. `metadata_get` falsy-fix — verified correct

Reviewed all 22 call sites. Every external caller uses the `(value := metadata_get(...)) is not None` pattern or doesn't care about falsy. The only old call that relied on truthiness semantics — the inline `pattern` lookup in `PrimitiveType.validate` — is migrated to the new API in this PR. Safe. Tests for `minimum=0` and `min_length=0` lock in the regression.

Minor: with the new semantics, `Metadata(config_name=None)` returns `default` rather than `None`, so you can't use it to opt out of an inherited `config_name`. Probably the right call but worth a one-line docstring note.

### 5. AST docstring extraction

- `inspect.getsource` fallback (`except OSError/TypeError: return {}`) handles frozen apps, REPL classes, and zipapps cleanly — schema generation just omits descriptions in those cases. Good.
- Per-class cache uses `cls.__dict__` lookup (not `hasattr`) to avoid inherited cache pollution — correct, and the comment explaining it is on point.
- MRO walk via `reversed(cls.__mro__)` with `merged.update` is right: ancestors fill first, subclass overrides last.
- Tested in `test_inherited_field_resolves_to_parent_docstring`.

### 6. Test coverage

- **Comprehensiveness tests are excellent** — every registered `PoeTask.TaskOptions`, `ExecutorOptions`, three `ConfigOptions` classes, `ArgSpec`, and 5 TypedDicts. Adding a new option without a docstring will fail CI. Strong guarantee.
- **Constraint tests are mostly synthetic.** Only `min_length` on `ShellTask.TaskOptions.interpreter` is exercised through real code. `pattern`/`minimum`/`maximum`/`max_length` only have synthetic `Annotated[...]` test classes. Acceptable since no production field uses those constraints yet, but worth flagging — and the missing `str | list[str]` + `min_length` union case from §1.

### 7. Performance & conventions

- Lazy imports are correctly in place. `_docstrings` is imported inside `description_for_field`; `ast`/`inspect`/`textwrap` never load during normal CLI startup. `re` is also imported lazily inside `PrimitiveType.validate`. Verified.
- Validation overhead is microseconds per field — invisible.
- `poe lint` and `poe types` clean.
- Walrus / no-single-char / docstring style: all clean.

### Summary

Approve the direction. The single thing worth addressing before this merges (and before Phase 2 builds on it): **make the union-metadata propagation either field-selective or test-pinned**, so the JSON Schema generator in Phase 2 can rely on `Metadata(min_length=N)` having one unambiguous meaning. Everything else is either minor polish or a future-proofing nice-to-have.

---

poethepoet/config/partition.py
```diff
@@ -186,21 +230,90 @@ class ConfigOptions(PoeOptions):
...
+        Provide one or more env files to be loaded before running tasks. If an
+        array is provided, files will be loaded in the given order.
+        """
+
         executor: Mapping[str, str | Sequence[str] | bool] | str = MappingProxyType(
             {"type": "auto"}
         )
+        """
+        Configure the executor type for running tasks. Can be 'auto', 'poetry',
+        'uv', 'virtualenv', or 'simple', with 'auto' being the default.
```
nat-n:
```suggestion
        Configure the executor for running tasks. The type can be 'auto', 'poetry',
        'uv', 'virtualenv', or 'simple', with 'auto' being the default. Some executor 
        types accept additional configuration options.
```

PR #391, opened 2026-05-20
Fix executor config validation to reject unknown keys

nat-n:
## Description of changes

This PR fixes a regression in executor configuration validation where unknown keys in `tool.poe.executor` were silently accepted instead of being rejected at config-parse time.

### The Problem

The `validate_config` method in `executor/base.py` was calling `ExecutorOptions.parse()` but not consuming the returned generator. Since generators are lazy, the validation logic inside `parse()` was never executed. This meant invalid configuration keys were only caught later (and wrapped in an error) when `PoeExecutor.get()` finally consumed the generator.

### The Fix

Changed line 396 in `executor/base.py` to consume the generator by calling `next()` on it:
```python
# Before
cls.__executor_types[executor_type].ExecutorOptions.parse(config)

# After
next(cls.__executor_types[executor_type].ExecutorOptions.parse(config))
```

This ensures validation happens immediately during config parsing, providing cleaner error messages to users.

### Testing

Added a regression test `test_global_executor_config_rejects_unknown_keys` that verifies:
- Unknown keys in executor config are rejected with a clear error message
- The error is raised during config validation (not wrapped later)
- The task execution fails cleanly with exit code 1

## Pre-merge Checklist

- [x] Bug fix is accompanied by a regression test
- [x] This PR targets the *development* branch

https://claude.ai/code/session_01Aza7YdEvFAh1Avci1j8pag

### DIFF
--- a/poethepoet/executor/base.py
+++ b/poethepoet/executor/base.py
@@ -393,7 +393,7 @@ class PoeExecutor(metaclass=MetaPoeExecutor):
             )
 
         else:
-            cls.__executor_types[executor_type].ExecutorOptions.parse(config)
+            next(cls.__executor_types[executor_type].ExecutorOptions.parse(config))
 
 
 def _stop_coverage():
--- a/tests/test_executors.py
+++ b/tests/test_executors.py
@@ -247,3 +247,32 @@ def test_global_executor_config_rejects_unknown_keys(temp_pyproject, run_poe):
     # would mean PoeExecutor.get() caught it instead of validate_config.
     assert "Couldn't parse executor options" not in result.capture
     assert result.stdout == ""
+
+
+def test_global_executor_config_rejects_wrong_value_type(temp_pyproject, run_poe):
+    """
+    The same generator-not-consumed bug that hid unknown keys also hid the
+    per-field value_type.validate(...) checks inside PoeOptions.parse. This
+    locks in that value-type errors on executor configs surface from
+    validate_config too (companion to #390).
+
+    Uses a bool value for `location` — accepted by the outer ConfigOptions
+    schema (which allows str/list[str]/bool for executor option values) but
+    invalid against the inner virtualenv ExecutorOptions schema (str | None).
+    Without the fix, the inner check is skipped and the error never surfaces.
+    """
+    project_path = temp_pyproject(
+        """
+            [tool.poe.executor]
+            type = "virtualenv"
+            location = true
+
+            [tool.poe.tasks]
+            greet = "poe_test_echo hi"
+        """
+    )
+    result = run_poe("-d", "greet", cwd=project_path)
+    assert result.code == 1
+    assert "Option 'location' must have a value of type" in result.capture
+    assert "Couldn't parse executor options" not in result.capture
+    assert result.stdout == ""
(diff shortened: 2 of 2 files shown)


nat-n (commented):
## Review

Tight, well-scoped bug fix. Root cause is correctly identified: `PoeOptions.parse()` is a generator (`yield` at line 130 of `options/base.py`) so calling it without iterating skips all validation entirely — not just the unknown-key check but also the missing-required-key check and the per-field `value_type.validate(...)` calls.

### Correctness ✅

The fix triggers validation as intended. `normalize()` yields exactly one item for a dict source, and the validation logic in `parse()` runs *before* the `yield`, so `next()` is sufficient to exercise the full validation pass for a single executor config.

### Suggestions

1. **`next()` is slightly brittle for an "exhaust for side effects" pattern.** It consumes only the first yielded item and would raise `StopIteration` (not `ConfigValidationError`) if the generator were ever empty. For the current single-dict usage this is fine, but a more defensive idiom would be:
   ```python
   for _ in cls.__executor_types[executor_type].ExecutorOptions.parse(config):
       pass
   ```
   or `collections.deque(..., maxlen=0)`. Marginal — feel free to ignore.

2. **Consider a dedicated "validate only" path.** `validate_config` discards the parsed result, so it's calling `parse()` purely for side effects. A small `PoeOptions.validate_only(config)` classmethod (or a `validate=True, construct=False` flag) would make the intent explicit and prevent this class of bug from recurring in other callers. Out of scope for this PR but worth a thought for the JSON Schema work landing nearby (#389/#392), which leans heavily on the declarative validation path.

### Issues to address

- **Base branch mismatch.** The checklist says "This PR targets the *development* branch" but the PR is actually opened against `main`. Per `CLAUDE.md`: *"Branch from `development` for features/bugfixes; Branch from `main` only for docs or hotfixes."* Unless this is being treated as a hotfix, please retarget to `development`.

- **Test could assert one more thing.** Since the same generator-not-consumed bug also masked *missing required key* errors (the `Missing required option` raise at `options/base.py:118` lives inside the loop body), a second tiny case asserting that a missing-required-key error is also raised from `validate_config` would lock in the full scope of the regression. Optional but cheap.

### Test coverage

Good. The "wrapped form would be 'Couldn't parse executor options …'" assertion is a nice belt-and-braces check that the error is raised by `validate_config` itself rather than the downstream `PoeExecutor.get()`. Comment explaining the regression is excellent.

### Performance / security

No concerns.

---

nat-n:
Thanks for the review.

- **Base branch**: retargeted to `development` (eba113f is the head; the PR base now points there).
- **Additional test case**: added `test_global_executor_config_rejects_wrong_value_type` covering the per-field `value_type.validate(...)` path that the same bug also masked. I went with value-type instead of missing-required-key because `type` is the only required executor field and is checked explicitly *outside* the generator at `executor/base.py:374`, so the missing-required raise inside the generator body isn't actually reachable from `validate_config`'s `parse()` call. The value-type case exercises the same "skipped because the generator was never iterated" failure mode and uses `location = true` on the virtualenv executor — accepted by the outer `ConfigOptions` schema (bool is permitted) but invalid against the inner `ExecutorOptions` schema (str | None).
- **`next()` brittleness** and **dedicated validate-only path**: skipped per your guidance ("feel free to ignore" and "out of scope"). Worth noting `next(...)` matches the established pattern at `executor/base.py:141` and the other `PoeOptions.parse` call sites — happy to refactor to `for _ in ...: pass` or extract a `validate_only` helper as a follow-up if you'd prefer.

---

Issue #390, opened 2026-05-20
PoeExecutor.validate_config silently skips validation (unconsumed generator)


nat-n:
## Summary

`PoeExecutor.validate_config` at `poethepoet/executor/base.py:427` calls `ExecutorOptions.parse(config)` but never iterates the returned generator, so all validation logic inside `PoeOptions.parse` is silently skipped. Root-level `tool.poe.executor` (and `include_script` executor) configs with unknown keys or invalid value types are accepted at parse time instead of being rejected.

Every other call site of `PoeOptions.parse` correctly wraps it in `next(...)` (see `executor/base.py:147`, `config/partition.py:190`, `task/base.py:309`). Line 427 is the lone outlier.

## Reproduction

```python
from poethepoet.executor.base import PoeExecutor

# Malformed: poetry's ExecutorOptions doesn't define `extra_bogus`
config = {"type": "poetry", "extra_bogus": "should be rejected"}

PoeExecutor.validate_config(config)
# Returns silently — no error raised.
```

For comparison, iterating the same `parse()` call directly does raise:

```python
from poethepoet.executor.poetry import PoetryExecutor

list(PoetryExecutor.ExecutorOptions.parse(config))
# raises ConfigValidationError: Unrecognized option 'extra_bogus'
```

## Blast radius

- Affects root-level `tool.poe.executor = {...}` configs and the `executor` option inside `include_script` items (both flow through `PoeExecutor.validate_config`).
- Does **not** affect task-level `executor` configs — those go through `TaskOptions.parse` which is correctly iterated.
- Specifically, unknown keys, value-type errors, and any future required-field violations on the executor option are silently accepted at the root level.

## Root cause

`PoeOptions.parse` is a generator function (returns `Iterator[PoeOptions]`). All validation — unknown-key checks, required-field checks, value-type validation — happens inside the generator body during iteration. Calling `parse(config)` without iterating just constructs a generator object; none of the body runs.

```python
# poethepoet/executor/base.py, validate_config (~line 403)
@classmethod
def validate_config(cls, config: dict[str, Any]):
    ...
    elif executor_type not in cls.__executor_types:
        raise ConfigValidationError(...)
    else:
        cls.__executor_types[executor_type].ExecutorOptions.parse(config)
        # ↑ unconsumed generator; nothing runs.
```

## Fix

One-line change at line 427 — wrap in `next(...)` to advance the generator (since the input is a single dict, `parse()` yields exactly one element):

```diff
[...]

PR #392, opened 2026-05-20
feat: Add task to generate JSON Schema based on PoeOptions

nat-n:
Stacked on #389 (Phase 1). Delivers the schema generator and lifecycle integration that depend on the foundations there.

Design spec: `docs/superpowers/specs/2026-05-17-poe-jsonschema-generation-design.md`
Implementation plan: `docs/superpowers/plans/2026-05-18-poe-jsonschema-phase2.md`

## Summary

**Phase 2 — Schema generator and parity tests:**

- **`poethepoet/schema/` package** (new): translates `PoeOptions` definitions into a complete draft-07 JSON Schema for the `tool.poe` config block. Output committed at `docs/_static/partial-poe.json` (~81KB, 33 definitions). Package is never imported during normal CLI invocations — zero startup-time impact.
- **`__schema_fragment__` hook** on `PoeOptions` (with a default implementation), overridden on `PoeTask` (wraps `TaskOptions` with the discriminator) and on `SwitchTask` / `SequenceTask` / `ParallelTask` / `ArgSpec` for irregular shapes. Cross-cutting compositions (task_def union with forward-compat fallback, executor tagged union, env/envfile/args polymorphism, `tasks`/`groups` patternProperties) live in `poethepoet/schema/fragments.py`.
- **`additionalProperties: false`** everywhere structural. Single deliberate exception: a forward-compat fallback branch matches task dicts that don't contain any known task-type discriminator — editors won't false-positive on future task types added in newer poe versions than the user's schema.
- **Approach B (full inlining)** instead of `allOf` + delta — draft-07's `additionalProperties: false` doesn't compose through `allOf`, so each task variant inlines its full property set (own + inherited). Trade-off documented in the spec.
- **Public registry accessors:** `PoeTask.get_task_class(key)` and `PoeExecutor.get_executor_class(key) / get_executor_types()` — public read-side complements to the metaclass registrations, replacing what would otherwise need name-mangled `_PoeTask__task_types` access from the schema generator.
- **Parity test suite** under `tests/schema/`:
  - `test_meta.py` — generated schema is itself valid draft-07; no dangling `$ref`s.
  - `test_fixture_configs.py` — every `tests/fixtures/*_project/` config validates against the schema (28 pass, 1 xfail for unimplemented `conditionals` feature).
  - `test_invalid_corpus.py` — seven curated invalid configs rejected by both runtime and schema.
  - `test_mutation.py` — parametrized (seed × mutator × path) cases, with `pytest.mark.xfail(strict=True)` for known runtime-only divergences tracked in `KNOWN_RUNTIME_ONLY_MISMATCHES`. 16 entries documented with spec §7 reasons.
- **Three legitimate schema-too-strict bugs** caught and fixed during parity testing: switch `case` accepts int/bool, `include` mixed-list polymorphism, task-level executor without `type` (via new `executor_task_option` variant).
- **One runtime bug surfaced** (filed as #390, NOT fixed here): `PoeExecutor.validate_config` silently skips validation due to an unconsumed generator. Schema correctly rejects what runtime silently accepts; the mutation test for this case is in `KNOWN_RUNTIME_ONLY_MISMATCHES` with `strict=True` so it'll surface as XPASS once #390 is fixed, prompting cleanup of the allowlist entry.

**Phase 3 — Lifecycle integration:**

- **`poe build-schema`** — regenerates `docs/_static/partial-poe.json`.
- **`poe check-schema-drift`** — runs `build-schema` and asserts `git diff --exit-code` is clean. Used by CI.
- **`poe test-schema`** — runs the parity suite (`pytest -m schema tests/schema/`).
- **`poe test`** and **`poe test-quick`** now exclude the `schema` marker (parity tests live in `test-schema` only).
- **`poe check`** sequence: docs-check → style → types → lint → check-schema-drift → test → test-schema.
- **CI workflow:** `code-quality` job adds "Check schema is up to date" and "Run schema parity tests" steps. `run-tests` job excludes the `schema` marker so parity tests don't redundantly run on every OS × Python matrix combination.

## Test plan

- [x] `poe test` — 1112 passed, 3 skipped, 95 deselected (the schema-marked tests, correctly excluded).
- [x] `poe test-schema` — 78 passed, 17 xfailed (all xfails documented in `KNOWN_RUNTIME_ONLY_MISMATCHES` or in the fixture xfail dict, citing spec §7).
- [x] `poe build-schema` — writes 81583 bytes deterministically.
- [x] `poe check-schema-drift` — clean (no drift between committed schema and what the generator would produce now).
- [x] `poe types`, `poe lint` — clean.
- [ ] Reviewer to verify the generated `docs/_static/partial-poe.json` looks reasonable — open in an editor with JSON Schema support, confirm intellisense for `tool.poe` keys.

## Manual follow-ups (out of scope, your domain)

- Submit a PR to [schemastore/schemastore](https://github.com/SchemaStore/schemastore) replacing `partial-poe.json` with the contents of `docs/_static/partial-poe.json` after a poe release.
- Optionally wire `poe build-schema` into the existing `bump-version` task as a release-time safety net (already enforced by CI between releases).
- Fix issue #390 (`PoeExecutor.validate_config` unconsumed generator) — separate small PR; will let one of the `KNOWN_RUNTIME_ONLY_MISMATCHES` entries be removed.


nat-n (commented):
## Review — JSON Schema Phase 2 + 3

Solid PR. The three-way split (`translate.py` for per-annotation, `fragments.py` for cross-cutting, `generator.py` for orchestration) is clean and well-justified. `__schema_fragment__` as an extension hook — co-locating irregular shapes with their owning classes — is the right design. The lifecycle gates (`check-schema-drift` in CI, `xfail(strict=True)` on the allowlist) will catch regressions cleanly. Zero-startup-cost claim verified: no non-schema module imports `poethepoet.schema` at top level, and the two `__schema_fragment__` call sites are never reached on a normal `poe <task>` invocation.

A handful of real schema gaps and two runtime-semantics changes worth flagging.

### Schema gaps — under-strict in places it shouldn't be

1. **`task_group.executor` and `include_script.executor` are not discriminated** (`partial-poe.json` lines 2376-2402 and 2514-2524). At runtime, `PoeExecutor.validate_config` runs on `include_script.executor` (`partition.py:422-424`) and group executors are similarly constrained — but the schema emits `{anyOf: [string, object]}` with no tagged-union structure. Either extend `_apply_task_property_refs` to swap these like the per-task case, or tighten the underlying TypedDict annotations. The former is the cheaper fix.

2. **Nested `tasks` inside a group are unvalidated.** `TaskGroup.tasks: Mapping[str, Any]` produces `{type: object}` (line 2407-2410) — no `patternProperties`, no recursion into `task_def`. The top-level `tasks_map` is rigorous; nested-in-group tasks bypass all of it. This is the biggest editor-UX gap.

3. **Switch `case` rejects `float`.** Schema scalar is `string | integer | boolean`; runtime `task/switch.py:111-114` just calls `str(value)` so `case = 1.5` would work. Not exercised by current mutators because no seed contains a switch task — see §test coverage below.

4. **`executor_partial` accepts any object without `type`** (line 554-561). `{executor: {garbage: "value"}}` at task level slips through. Matches runtime, so not a parity violation, but notable permissiveness worth a follow-up.

### Runtime-semantics changes in `partition.py` worth a CHANGELOG note

The PR calls these "schema-too-strict fixes" but two of them widen/narrow runtime acceptance:

- **`_GROUP_NAME_PATTERN` is now ASCII-only.** Previously `re.fullmatch(r"[\w\-_]+", ...)` matched Unicode letters; now it doesn't. Previously-valid Unicode group names are now rejected. Test at `test_options.py:281-292` locks in the rejection. Probably the right call (matches schema convention) but it's a minor breaking change for any user with non-ASCII group names.

- **`include` accepts mixed lists.** Type changed from `str | Sequence[str] | Sequence[IncludeItem]` to `str | Sequence[str | IncludeItem]` (`partition.py:285,482`). Mixed lists like `["a.toml", {"path": "b.toml"}]` were previously rejected (failed both Sequence-union branches), now accepted. No new test covers the mixed-list runtime case — please add one.

- Task-level `executor_partial` inheritance is a pure schema-side fix matching existing runtime — well-targeted.

### `KNOWN_RUNTIME_ONLY_MISMATCHES` — one entry is hiding a real runtime bug

15 of 16 entries are legitimately "schema correctly stricter than runtime first-pass parse, runtime catches it at task-load time" — fine.

**The executors entry** (`test_mutation.py:143-148`) is the runtime bug from #390: `PoeExecutor.validate_config` calls `cls.__executor_types[...].ExecutorOptions.parse(config)` and discards the generator. **This is being fixed in #391** — once that merges, this allowlist entry should XPASS-and-fail-strict, prompting cleanup. Worth a cross-reference in the test comment.

### Test coverage — mutation matrix has a switch-shaped hole

The seed quality matters because the mutation matrix is exhaustive-over-seeds, not random fuzz. Current seeds (`simple.toml`, `executors.toml`, `complex.toml`) cover cmd/script/sequence/shell + deps but contain **no `switch` task**. All switch corner cases (e.g. `case = 1.5`, mixed-discriminator dicts) are untested by mutation. Adding a `switch.toml` seed is low-effort and would lift coverage substantially.

`xfail(strict=True)` discipline is good — when a schema fix lands, the xfail becomes XPASS and prompts allowlist cleanup. Lifecycle gates work: I confirmed `check-schema-drift` exits 1 with a visible diff when an annotation docstring is modified, exits 0 when reverted.

### Smaller items

- **Naming inconsistency:** `PoeTask.get_task_class` / `get_task_types` / `lookup_task_spec_cls` — three verbs on the same class for the same kind of lookup. Worth aligning in a follow-up (not blocking).

- **`_apply_task_property_refs`** (`generator.py:130-152`) silently `continue`s when `f"{key}_task"` isn't in `existing`. If a future task forgets to register via `task_def_schema`, that bug is masked. Suggest `assert def_name in existing` for the `_task` variant.

- **`translate.py:72-95` reaches into `_annotation` / `_values` / `_value_type` / `_schema` / `_optional_keys`** — acknowledged in the module docstring as intentional coupling, but it means any `TypeAnnotation` refactor must update both `validate()` and `translate()`. Worth a brief note in CHANGELOG.

- **`SchemaContext.definitions`** returns a fresh dict per access (`context.py:30-38`) — defensive against mutation but O(n) per call, and `fragments.py:324` accesses it inside a loop. Fine at current ~70 defs; consider `MappingProxyType` for constant-time read-only access at future scale.

- **`test_schema_cli.py:25-32`** spawns a subprocess without `env=` — fine on dev box, could inherit surprising state on CI with coverage instrumentation. Minor.

- **`_wrap_with_metadata` mutation concern from #389** doesn't recur here, but the `mutate_definition` post-hoc property swap in `_apply_task_property_refs` is a slightly fragile pattern — if a future task adds a non-standard `args`/`env`/`envfile`/`executor` shape, the swap silently clobbers it.

### Summary

The actionable items, in priority order:
1. **Tighten `task_group.executor` and `include_script.executor` to use the discriminated executor union.**
2. **Recurse `task_def` into nested `tasks` inside groups** — currently the biggest editor-UX gap.
3. **Add a `switch.toml` seed** to extend mutation coverage to switch corner cases.
4. **CHANGELOG note** for the group-name ASCII narrowing and the `include` mixed-list widening; add a runtime test for the latter.
5. **Cross-reference #391** in the executors `KNOWN_RUNTIME_ONLY_MISMATCHES` comment.

Everything else is polish or a future-proofing nice-to-have. Approve the direction; the gaps above are addressable without restructuring.

---

PR #400, opened 2026-06-06
feat: Add `buffer` output_mode for parallel tasks

### LINKED ISSUE #361: Print output from parallel tasks in one batch instead of line-by-line
Hi! I'm migrating from tox to poe, and one thing that I can't figure out is if it's possible to capture task's output and print it in one batch after it finishes. I see that I can save output to a file, but that's not exactly what I'm looking for.

nat-n:
## Summary
- add `output_mode = "buffer"` for `parallel` tasks: buffer each leaf task's stdout and flush its prefixed output on completion, failure, or cancellation
- loosely bound the memory used to assemble subtask output (~4 MiB per subtask) and wrap any longer line with an inserted break (warned in verbose mode), so a sibling's output is never spliced into the middle of a line
- add `output_mode` to the JSON schema, and document the mode — plus redirecting a subtask's output to a file — in the docs and bundled skill
- add regression coverage for wrapping, sibling-corruption, the wrap warning, and the cancellation flush

## Notes
- buffered output is scoped to leaf-task stdout, matching the current parallel output model
- isolate test caches under `tests/temp` and assert venv installs succeed

Closes #361

### DIFF
--- a/docs/_static/partial-poe.json
+++ b/docs/_static/partial-poe.json
@@ -857,6 +857,12 @@
           "description": "If set, the parallel task will continue running even if one of the subtasks\nfails.",
           "enum": [true, false, "return_zero", "return_non_zero"]
         },
+        "output_mode": {
+          "default": "stream",
+          "description": "Controls how leaf-task stdout is emitted: either streamed line-by-line\nor buffered output on task completion.",
+          "enum": ["stream", "buffer"],
+          "type": "string"
+        },
         "parallel": {
           "description": "Runs an array of subtasks concurrently. Each subtask runs in its own\nsubprocess; output lines are interleaved and prefixed with the subtask\nname by default.",
           "items": {
@@ -952,6 +958,12 @@
           "default": false,
           "description": "If set, the parallel task will continue running even if one of the subtasks\nfails.",
           "enum": [true, false, "return_zero", "return_non_zero"]
+        },
+        "output_mode": {
+          "default": "stream",
+          "description": "Controls how leaf-task stdout is emitted: either streamed line-by-line\nor buffered output on task completion.",
+          "enum": ["stream", "buffer"],
+          "type": "string"
         },
         "parallel": {
           "description": "Runs an array of subtasks concurrently. Each subtask runs in its own\nsubprocess; output lines are interleaved and prefixed with the subtask\nname by default.",
--- a/docs/tasks/task_types/parallel.rst
+++ b/docs/tasks/task_types/parallel.rst
@@ -10,7 +10,7 @@ A **Parallel task** is defined by an array of other tasks to be run concurrently
 
 By default the contents of the array are interpreted as references to other tasks (actually an inline :doc:`ref<ref>` task). However, this behaviour can be altered by setting the :toml:`default_item_type` option locally on the parallel task.
 
-Subtask outputs are streamed to the console as they arrive with a prefix identifying the task.
+Subtask outputs are streamed to the console by default as they arrive with a prefix identifying the task.
 
 
 Available task options
@@ -25,6 +25,9 @@ The following options are also accepted:
 
 **default_item_type** : ``str`` :ref:`📖<Changing the default item type>`
   Change the task type that is applied to string array items in this parallel group.
+
+**output_mode** : ``Literal["stream", "buffer"]`` :ref:`📖<Buffer subtask output>`
+  Controls how subtask stdout is rendered. The default ``stream`` mode prints each line as it arrives, while ``buffer`` keeps each leaf task's stdout grouped together and prints it later with the usual prefixes.
 
 **prefix** : ``str`` :ref:`📖<Customize output prefixing>`
   Set the prefix applied to each line of output from subtasks. By default this is the task name.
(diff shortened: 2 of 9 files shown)


nat-n:
## Code Review

### Core logic — `poethepoet/task/parallel.py`

**Module-level constant evaluated at import time**

```python
BUFFERED_STDOUT_LIMIT = _get_buffered_stdout_limit()
```

`POE_TEST_BUFFERED_STDOUT_LIMIT` is read once at module import. This works correctly for tests (subprocess gets a fresh import), but the behaviour is non-obvious. A short comment explaining this is intentional (and test-only) would help future readers.

---

**Buffer overflow logic is hard to follow**

The two-condition structure in the buffered path:

```python
if buffered_size and buffered_size + len(line) > BUFFERED_STDOUT_LIMIT:
    # flush, reset

buffered_lines.append(line)
buffered_size += len(line)

if buffered_size >= BUFFERED_STDOUT_LIMIT and len(buffered_lines) == 1:
    # flush, reset
```

The first condition's `and buffered_size` (falsy-check) intentionally skips flushing when the buffer is empty, leaving the oversized-first-line case to the second check. This reads like a bug at first glance. A comment explaining that the second branch handles "single oversized line into an otherwise empty buffer" would help considerably.

---

**`_iter_output_lines` missing type annotations**

```python
async def _iter_output_lines(self, stdout):
```

The rest of the codebase uses type hints throughout. `stdout` should be typed (likely `asyncio.StreamReader` or the relevant `PoeProcess` attribute type) and the return type should be declared as `AsyncIterator[bytes]`.

---

**Stream-mode behavior change not called out**

The PR silently changes how stream mode reads output — from `readline()` to chunk-based reads via `_iter_output_lines`. This is generally better (handles lines without trailing newlines, fewer syscalls), but it's a behavioral change to existing functionality not mentioned in the PR description and not tested in isolation. Worth noting explicitly.

---

**`write` + `flush` is still non-atomic across concurrent tasks**

Even in buffer mode, two tasks completing near-simultaneously can have their `write(b"".join(...))` + `flush()` calls interleaved at the OS level since there's no lock. The docs say output is "held back until the leaf task exits" — true per-task, but this doesn't guarantee blocks won't interleave if tasks finish simultaneously. This is a pre-existing limitation inherited from stream mode, but it's worth a note since buffer mode creates an expectation of non-interleaved output.

---

### Test infrastructure — `tests/conftest.py`

**`_get_poetry_tool_site_packages` is Unix-specific**

The shebang heuristic (`#!`) and `lib/pythonX.Y/site-packages` path structure are Unix conventions. On Windows the layout is `Lib/site-packages` with no version component. The function returns `None` on failure so it degrades gracefully, but the intended benefit won't materialise on Windows CI.

---

**`--no-deps` in `install_into_virtualenv` needs justification**

```python
"--no-build-isolation",
"--no-deps",
```

Skipping dependency installation could cause silent runtime failures in tests rather than install-time failures, making them harder to diagnose. A comment confirming these test fixtures have no meaningful transitive deps would make this safe to leave as-is.

---

### Tests — `tests/test_parallel_tasks.py`

**`BUFFER_LIMIT` is defined but never used**

```python
BUFFER_LIMIT = 4 * 1024 * 1024  # never referenced
```

Dead code — only `BUFFER_LIMIT_OVERRIDE_ENV` is used in tests.

---

**`format_parallel_prefix` duplicates implementation logic**

This helper replicates the truncation-at-16 + `…` logic from `_format_output_lines`. If the `prefix_max` default ever changes, this silently diverges. At minimum, sharing a constant (e.g. exporting `DEFAULT_PREFIX_MAX` from `parallel.py`) would keep them in sync.

---

**No test for the cancellation path**

The docs and PR summary say output is flushed "on completion, failure, or cancellation" but only the failure case has test coverage. A cancellation test would complete the contract.

---

**Missing edge case: `prefix=False` + `output_mode="buffer"`**

`_flush_output_buffer` handles `prefix=b""` correctly, but there's no test that exercises `output_mode="buffer"` combined with `prefix = false`.

---

### `tests/test_cli.py`

**Unrelated change bundled in**

The switch of `test_pass_dry_run_and_verbosity_to_script` from `run_poe` to `run_poe_subproc` appears to be an environment compatibility fix unrelated to buffered output. Bundling it here makes the diff harder to review and harder to revert independently.

---

### Summary

| File | Finding | Severity |
|---|---|---|
| `parallel.py` | `_iter_output_lines` missing type annotations | Minor |
| `parallel.py` | Buffer overflow two-condition logic needs a comment | Minor |
| `parallel.py` | Stream-mode behaviour change not documented | Minor |
| `parallel.py` | Concurrent writes still non-atomic (pre-existing) | Info |
| `conftest.py` | `--no-deps` needs justification comment | Minor |
| `conftest.py` | `_get_poetry_tool_site_packages` Unix-only | Minor |
| `test_parallel_tasks.py` | `BUFFER_LIMIT` unused | Minor |
| `test_parallel_tasks.py` | `format_parallel_prefix` duplicates prefix logic | Minor |
| `test_parallel_tasks.py` | Missing cancellation test | Minor |
| `test_parallel_tasks.py` | Missing `prefix=False` + buffer mode test | Minor |
| `test_cli.py` | Unrelated change bundled in | Minor |

The core feature is well-designed and the main cases are covered. No blocking issues — mostly polish and test coverage gaps.

---

PR #402, opened 2026-06-07
fix(args): support all three CLI styles for multi-value args

nat-n:
## Summary

Make `multiple` args behave predictably across every CLI style, and always resolve them to a list.

Before this PR, `multiple` args had three sharp edges:

- Repeated flags silently dropped all but the last value: `--foo a --foo b` → `["b"]`.
- `multiple = N` enforced the count *per occurrence*, rejecting valid splits like `--widgets a --widgets b`.
- A scalar `default` crashed (`'str' object has no attribute 'extend'`) or got silently prepended onto supplied values once `action="extend"` was in play.

## Behavior

- Accept all three styles, freely mixed: `--engines v2 v8`, `--engines v2 --engines v8`, `--engines v2 v8 --engines v10`.
- Require the *total* value count to equal N for `multiple = N`; report mismatches in argparse's own style (`expected 2 values, got 3`).
- Always surface a `multiple` arg as a list: omitted → `[]`, or `[default]` when a default is set.
- Replace the default with supplied values — never extend it.

## Implementation

`poethepoet/task/args.py`:

- Switch `multiple = true` and `multiple = N` to `action="extend"`; parse `multiple = N` as `nargs="*"` and check the total post-parse (`_validate_exact_count`), since argparse can't express a total across occurrences.
- Keep argparse's default empty for `multiple` args; apply the configured default after parsing (`_normalize_multiple_defaults`) so `extend` can't prepend it onto user values.

## Behavior change

- Omitting a `multiple` arg now yields `[]` instead of `None` (option args matched what argparse already did for `*` positionals).
- Invisible to `cmd`/`shell`/`expr`/`switch` — env exposure already treats `None`, `[]`, and `False` as unset.
- `script`/`expr` tasks receive `[]` rather than `None`, matching the documented "list of values" contract.

## Tests

New (TDD):

- `test_script_with_repeated_flag_accumulates`, `test_script_with_mixed_multi_styles` — repeated/mixed styles accumulate.
- `test_script_with_exact_count_across_occurrences`, `test_script_exact_count_wrong_total_errors` — `multiple = N` validates the total.
- `test_multiple_args_default_to_empty_or_scalar` — omitted args resolve to `[]` / `[default]`.
- `test_multiple_arg_scalar_default_is_overridden_not_extended` — supplied values replace the default, not extend it.

Updated: the `widgets` failure-case and `None`→`[]` assertions across `test_script_tasks`, `test_expr_task`, `test_ref_task`, plus the task listing in `test_cli`.

## Docs

- `docs/guides/args_guide.rst` and the bundled `poethepoet/skills/poethepoet/references/args-reference.md` — name the three styles, total-count semantics, and the always-a-list default behavior.

## Test plan

- [x] `poe lint` — clean
- [x] `poe types` — clean
- [x] `poe format` — clean
- [x] `poe test` (full suite) — 1252 passed, 3 skipped

## Notes

Fixes the bug surfaced in #401 (the agent-skill PR), where `poe eval-skill --eval 6 --eval 7` ran only eval 7. The harness can now use the natural repeated-flag form.


poethepoet/task/args.py
```diff
@@ -432,12 +432,20 @@ def _get_argument_params(self, arg: ArgSpec, env: TaskEnv):
...
         if multiple is True:
-            if required:
-                result["nargs"] = "+"
-            else:
-                result["nargs"] = "*"
+            # action="extend" so each repeated occurrence of the flag extends the
+            # accumulator (rather than overwriting it as the default action would).
+            # This lets the three CLI styles — space-separated values, repeated
+            # flag, and a mix — all work, e.g. `--foo a b`, `--foo a --foo b`, or
+            # `--foo a --foo b c`.
```
nat-n:
```suggestion
```

poethepoet/task/args.py
```diff
@@ -432,12 +432,20 @@ def _get_argument_params(self, arg: ArgSpec, env: TaskEnv):
...
+            result["nargs"] = "+" if required else "*"
+            result["action"] = "extend"
         elif multiple and isinstance(multiple, int):
-            result["nargs"] = multiple
+            # For an exact-count multi (multiple = N), accept any combination of
+            # styles as long as the total count equals N. argparse can't express
+            # "exactly N total" natively, so use nargs="*" + extend here and
+            # validate the total post-parse in `parse()`.
```
nat-n:
```suggestion
```

poethepoet/task/args.py
```diff
@@ -432,12 +432,18 @@ def _get_argument_params(self, arg: ArgSpec, env: TaskEnv):
...
+            # accumulator (rather than overwriting it as the default action would).
+            # This lets the three CLI styles — space-separated values, repeated
+            # flag, and a mix — all work, e.g. `--foo a b`, `--foo a --foo b`, or
+            result["nargs"] = "+" if required else "*"
+            result["action"] = "extend"
         elif multiple and isinstance(multiple, int):
-            result["nargs"] = multiple
+            # For an exact-count multi (multiple = N), accept any combination of
+            # styles as long as the total count equals N. argparse can't express
+            # "exactly N total" natively, so use nargs="*" + extend here and
```
nat-n:
```suggestion
```

poethepoet/task/args.py
```diff
@@ -432,12 +432,18 @@ def _get_argument_params(self, arg: ArgSpec, env: TaskEnv):
...
 
         if multiple is True:
-            if required:
-                result["nargs"] = "+"
-            else:
-                result["nargs"] = "*"
+            # action="extend" so each repeated occurrence of the flag extends the
+            # accumulator (rather than overwriting it as the default action would).
+            # This lets the three CLI styles — space-separated values, repeated
+            # flag, and a mix — all work, e.g. `--foo a b`, `--foo a --foo b`, or
```
nat-n:
```suggestion
```

PR #405, opened 2026-06-24
fix(evals): detect counter-example labels on preceding lines

nat-n:
Fix a false-fail in the skill eval grader, surfaced by the latest eval run (eval-6 scored *below* the no-skill baseline despite a correct answer).

- Scan a small window of preceding lines for a counter-example marker, not just the matched line — labels like `# Wrong` typically sit a line or two *above* the offending value (the usual ✅/❌ contrast block)
- Leave the line-start `found` matching unchanged, so eval-7's negate assertion is unaffected
- Add regression tests for the grading heuristics (none existed)

Does not touch the skill or its content; this is eval-harness only.

### DIFF
--- a/tests/skills/run_evals.py
+++ b/tests/skills/run_evals.py
@@ -158,18 +158,29 @@ _COUNTER_EXAMPLE_MARKERS = (
     "not recommended",
 )
 
+# How many preceding lines to scan (in addition to the matched line) when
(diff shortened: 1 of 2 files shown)


nat-n:
Folding this into #404 — the grading fix fits with the skill improvement work. Same commit is now on that branch.

PR #407, opened 2026-06-28
feat: Add `uses_env` task option for loading variables from a task's output

nat-n:
## Description of changes

Adds a new `uses_env` task option — a sibling to `uses` for loading **multiple** environment variables produced by a single upstream task.

Where `uses` captures a task's stdout into one named variable, `uses_env` captures the output of one or more tasks and parses each as an env file (dotenv syntax) to load environment variables from it — zero or more per task, which the task names itself — into the host task's environment.

The motivating use case is loading several related variables emitted together by one command, such as all the `AWS_*` credentials from a tool like `aws-vault`:

```toml
[tool.poe.tasks._aws-creds]
cmd = "aws-vault export my-profile"

[tool.poe.tasks.deploy]
cmd = "terraform apply"
uses_env = "_aws-creds"
```

It accepts a single task invocation or a list (applied in order), supports the same task references and argument passing as `uses`/`deps`, and keeps `_`-prefixed values private to the subprocess just like `uses`. Misuse is reported with clear errors — referencing a `capture_stdout`/`use_exec` task, declaring it on a `switch` case, or a producer whose output isn't valid env-file syntax.

The user docs (options reference + composition guide), the bundled agent skill, and the generated JSON schema are all updated to match.

Two incidental changes ride along: a fix for a schema-assertion test that was already failing on `main`, and clearer environment-setup guidance in `CLAUDE.md`.

## Additional hardening (landed during review)

Reviewing this branch surfaced a few related issues, fixed here with regression tests:

- **Reliable capture of `uses_env` output.** A `switch` task's output was whitespace-collapsed before `uses_env` could parse it as an env file, and a task referenced by *both* `deps` and `uses_env` could be deduplicated in the execution graph as uncaptured (raising a `KeyError`). Both are fixed so `uses_env` reliably receives raw upstream output.

- **Curated errors for capturing a task type that can't be captured.** `sequence`/`parallel` tasks can't have their output captured; referencing one via `uses`, `uses_env`, or a capturing `ref` previously either slipped through validation or crashed with an unhandled `AssertionError`. It now produces a clear config-time error — including when capture is forwarded transitively through a `ref` or `switch`. This is implemented with:
  - a `Disinherited[T]` option-annotation marker, so a task type can drop an inherited option from its config surface (one declaration that both the generated schema and the config parser honour via `get_fields`), replacing three imperative rejection sites on `sequence`/`parallel`; and
  - a polymorphic `TaskSpec.accepts_option`, which forwarding task types (`ref`, `switch`) override to recurse into their target(s), so the check is complete and correct at config time.

- **Memoization fix.** `PoeOptions.get_fields`/`get_field_attribute` had a presence check that compared against an unmangled name literal and so never matched the stored key — the caches recomputed on every call. They are now correctly memoized per class.

## Pre-merge Checklist

- [x] Bug fixes (if any) are accompanied by a test (if not too complicated to do so)

https://claude.ai/code/session_019sWh9vZKTtzqUVt6Ft7aUR


poethepoet/task/base.py
```diff
@@ -173,6 +173,24 @@ def from_task(cls, parent_task: PoeTask, task_spec: PoeTask.TaskSpec):
         )
 
 
+def _parse_uses_env_output(
```
nat-n: better to just inline this logic in the one call sight than have another top level function in this module. The import should be the next line after `if uses_env_outputs:`

poethepoet/task/base.py
```diff
@@ -234,6 +252,14 @@ class TaskOptions(PoeOptions):
         accessible in this task.
         """
 
+        uses_env: str | Sequence[str] = ()
+        """
+        Allows this task to capture the output of other tasks which are executed
+        first, and load environment variables from it. Each referenced task's
+        stdout is parsed as an env file (dotenv syntax), so a single task can
+        provide zero or more variables, which it names itself.
```
nat-n:
```suggestion
        Makes this task depend on other tasks, and run with environment variables parsed
        from their captured output. Each referenced task's stdout is parsed as an 
        env file (dotenv syntax), so a single task can provide zero or more variables, 
        which it names itself.
```

poethepoet/context.py
```diff
@@ -178,14 +180,29 @@ def _get_dep_values(
             for var_name, invocation in used_task_invocations.items()
         }
 
+    def _get_dep_env_outputs(
```
nat-n: Lets think about whether it would actually be better to do the envfile parsing around here at the context level and then just vend the variables? could even handle both uses and uses_env together with a single method call?

poethepoet/task/switch.py
```diff
@@ -17,7 +17,7 @@
 
 
 DEFAULT_CASE = "__default__"
-SUBTASK_OPTIONS_BLOCKLIST = ("args", "uses", "deps")
+SUBTASK_OPTIONS_BLOCKLIST = ("args", "uses", "uses_env", "deps")
```
nat-n: sort?

poethepoet/skills/poethepoet/references/task-options.md
```diff
@@ -128,6 +128,27 @@ uses = { _version = "_get-version" }
 
 ---
 
+## uses_env
+
+Capture the stdout of other task(s) and load environment variables from it, parsing each one's output as an env file (dotenv syntax). Unlike `uses`, a single subtask can provide zero or more variables, which it names itself.
+
+```toml
+[tool.poe.tasks._aws-creds]
+shell = "aws-vault exec my-profile -- env | grep '^AWS_'"
```
nat-n: Check the docs on aws-vault, I believe there's a more idiomatic approach supported

poethepoet/context.py
nat-n: I looked at moving this to the context level but kept the parsing in `get_task_env`. The reason: env file values support `${VAR}` expansion, and that has to resolve against the task's assembled base env (`parent_env`) — which, for a `uses_env` on a sequence/parallel subtask, includes that subtask's own inherited env layer. That env is only available here in `get_task_env`; the context only has the root run env. So vending pre-parsed variables from the context would either drop that expansion or require threading the `TaskEnv` down into the context.

I did inline the parse and drop the helper (per the other comment). Happy to revisit unifying with `uses` if you'd rather push the base env down into the context — just wanted to flag the expansion constraint first.

---

poethepoet/skills/poethepoet/references/task-options.md
nat-n: Good call — confirmed from the aws-vault source: `aws-vault export <profile>` defaults to `--format=env` and prints `AWS_ACCESS_KEY_ID=…` lines straight to stdout, so a plain `cmd` does it with no `exec -- env | grep`. Updated the example to `cmd = "aws-vault export my-profile"`.

---

PR #409, opened 2026-06-30
fix(args): restore default for absent multiple positional args

nat-n:
## What

- An absent positional arg declared with `multiple` no longer drops its configured `default`.

## Why

- 0.47.0's multi-value arg rework applies defaults only when argparse yields `None`.
- argparse collapses an absent `nargs="*"` **positional** to `[]`, not `None`, so its default was silently skipped.
- e.g. `cmd = "mypy $files"` with `default = "src tests"` expanded to `mypy` with no target.

## Fix

- Treat a positional's `[]` as absent (a positional can't be given an explicit empty), restoring the default.
- Leave an explicitly-emptied option's `[]` intact.
- Guard `_validate_exact_count` so an absent `multiple = N` positional doesn't trip a spurious "expected N values, got 0".

## Tests

- New scripts fixture: positional `multiple` arg with a default; covers default-applied and explicit-override cases.
- New cmd-task coverage mirroring the reported break end-to-end (default resolves through env-var exposure and `$files` word-splitting): default reaches the command line, supplied values override, absent `multiple = N` resolves to empty rather than erroring.

## Unrelated

- Aligns a stale `print_result` schema-clause assertion with the generator output (`"type": "string"` added in `d2f9c11`), which was already failing on `development`.

### DIFF
--- a/poethepoet/task/args.py
+++ b/poethepoet/task/args.py
@@ -521,7 +521,7 @@ class PoeTaskArgs:
             # and the first entry of `options` for positionals (argparse default).
             key = arg.options[0] if arg.positional else arg.name
             value = parsed_args.get(key)
-            if value is None:
+            if value is None or (arg.positional and value == []):
                 continue
             if len(value) != multi:
                 parser.error(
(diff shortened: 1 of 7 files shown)


poethepoet/task/args.py
```diff
@@ -535,20 +535,16 @@ def _normalize_multiple_defaults(
...
         for arg in self._args:
             if not arg.multiple:
                 continue
             # dest is `arg.name` for option args and `arg.options[0]` for
             # positionals, matching `_validate_exact_count`.
             key = arg.options[0] if arg.positional else arg.name
-            if parsed_args.get(key) is not None:
+            value = parsed_args.get(key)
+            if value is not None and not (arg.positional and value == []):
+                # Position arg with multiple=True arrives as []
```
nat-n:
```suggestion
                # Omitted option arg with multiple=True arrives as None
                # Omitted positional arg with multiple=True arrives as []
```

